PR #26, opened 2015-01-02
Add method to get names for an edge.

dnesbitt61:
Protect against offsets exceeding the size of the names list.


dgearhart:
:+1:

PR #29, opened 2015-01-05
Return speed as int

dnesbitt61:
Will use "speed  cost factors" rather than dividing through by speed.  So don't want to return float

### DIFF
--- a/src/baldr/directededge.cc
+++ b/src/baldr/directededge.cc
@@ -35,6 +35,11 @@ uint8_t DirectedEdge::reverseaccess() const {
   return reverseaccess_.v;
 }
 
+// Gets the speed in KPH.
+uint8_t DirectedEdge::speed() const {
+  return speed_;
+}
+
 // TODO - methods for individual access
 
 bool DirectedEdge::ferry() const {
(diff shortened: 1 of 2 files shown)


src/baldr/directededge.cc
```diff
@@ -35,6 +35,11 @@ uint8_t DirectedEdge::reverseaccess() const {
   return reverseaccess_.v;
 }
 
+// Gets the speed in KPH.
+uint8_t DirectedEdge::speed() const {
```
kevinkreiser: when we start doing airplanes and bullet trains this wont be enough precision. seriously though doubling the width of this will have no impact on performance i'd wager

dnesbitt61: Planes and trains run on schedules so I don't think we will take speed into
account for those travel modes - just the time between stops based on the
schedule data. But we can increase this if needed later.

On Mon, Jan 5, 2015 at 9:56 PM, Kevin Kreiser <notifications@github.com>
wrote:

> In src/baldr/directededge.cc
> <https://github.com/valhalla/baldr/pull/29#discussion-diff-22505770>:
>
> > @@ -35,6 +35,11 @@ uint8_t DirectedEdge::reverseaccess() const {
> >    return reverseaccess_.v;
> >  }
> >
> > +// Gets the speed in KPH.
> > +uint8_t DirectedEdge::speed() const {
>
> when we start doing airplanes and bullet trains this wont be enough
> precision. seriously though doubling the width of this will have no impact
> on performance i'd wager
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/valhalla/baldr/pull/29/files#r22505770>.
>



-- 
David Nesbitt
dave@mapzen.com
443-206-1819

PR #31, opened 2015-01-06
removed the nodes from edgeinfo and reduced text offset sizes

dgearhart:


dnesbitt61:
ship it

PR #32, opened 2015-01-06
updates for ll class to be lon/lat

kevinkreiser:
reflecting changes in midgard pr: https://github.com/valhalla/midgard/pull/6


dnesbitt61:
ok, once tests pass

PR #36, opened 2015-01-07
edgeinfo to use compressed shape

kevinkreiser:
update edgeinfo to remove extra stuff and make use of encoded shape. shape will be cached once decoded the first time and wont be decoded until asked for


gknisely:
ship

dnesbitt61:
ok

PR #37, opened 2015-01-08
Add hierarchy transitions and shortcut flags. Change lanes to lanecount.

dnesbitt61:
To support hierarchies.


dgearhart:
:+1:

PR #38, opened 2015-01-08
Add virtual destructors 

dnesbitt61:
so the class is polymorphic and can cast to derived (builder) classes


kevinkreiser:
:ship:

PR #39, opened 2015-01-08
Fix best road class for a node.

dnesbitt61:
Use 3 bits and use RoadClass enum in the access methods. Add a method to test if a tile exists.

### DIFF
--- a/src/baldr/graphreader.cc
+++ b/src/baldr/graphreader.cc
@@ -3,6 +3,7 @@
 #include <string>
 #include <iostream>
 #include <fstream>
+//#include <boost/filesystem/operations.hpp>
 
 namespace valhalla {
 namespace baldr {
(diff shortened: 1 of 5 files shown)


src/baldr/graphreader.cc
```diff
@@ -14,6 +14,12 @@ GraphReader::GraphReader(const boost::property_tree::ptree& pt):GraphReader(Tile
 GraphReader::GraphReader(const TileHierarchy& th):tile_hierarchy_(th) {
 }
 
+// Method to test if tile exists
+bool GraphReader::DoesTileExist(const GraphId& graphid) const {
+  return std::ifstream(GraphTile::Filename(
```
kevinkreiser: won't this let the file hanging open? shouldnt you close it before returning?

kevinkreiser:
other than comment looks good

PR #40, opened 2015-01-09
lat, lon parse order

kevinkreiser:
so to match the code i changed the parsing of lat lon to be lon lat but that seems to be not a convention humans use to much. so i'm changing it back. this will change anything that parses strings into `PointLL` objects or `Location` objects.


dgearhart:
:+1:

PR #42, opened 2015-01-14
Add GetNames method that takes the edgeinfo shared ptr. Add comments.

dnesbitt61:
Often we have the shared ptr, so no need to get it again when retrieving names.


dgearhart:
:+1:

PR #43, opened 2015-01-15
tile uri generation to make sure not too many files in one dir

kevinkreiser:
this makes it so that tiles will be written only 1000 at a time to any single dir and that no dir has no more than 1000 other dirs in it. there is  a test against this as well.

it also does a little clean up of shared pointer stuff and the GetNames function

this breaks mjolnir and thor, prs in the hopper

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,6 +41,7 @@ test-driver
 test/location
 test/graphid
 test/tilehierarchy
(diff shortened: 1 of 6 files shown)


dnesbitt61:
ok. Let me know what changes need to be made in mjolnir

PR #44, opened 2015-01-16
Increase edge_index. 

dnesbitt61:
Create constants for max. edges per tile and max. edges per node.
Need to recreate data once this is merged.


kevinkreiser:
ship

PR #45, opened 2015-01-16
location and path location updates

kevinkreiser:
this adds the ability to set the point to which loki (for example) correlated the location when creating the pathlocation. previously this information wasn't available.


dnesbitt61:
ok

PR #47, opened 2015-01-20
log disk cache misses with debug only

kevinkreiser:
to see this you'd need to build baldr after configuring with CPPFLAGS="-DLOGGING_DEBUG"

### DIFF
--- a/src/baldr/graphtile.cc
+++ b/src/baldr/graphtile.cc
@@ -2,6 +2,7 @@
 #include <valhalla/midgard/tiles.h>
 #include <valhalla/midgard/aabb2.h>
 #include <valhalla/midgard/pointll.h>
+#include <valhalla/midgard/logging.h>
 
 #include <string>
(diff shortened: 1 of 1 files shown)


dgearhart:
:+1:

src/baldr/graphtile.cc
```diff
@@ -94,8 +95,9 @@ GraphTile::GraphTile(const TileHierarchy& hierarchy, const GraphId& graphid)
 
     // Set the size to indicate success
     size_ = filesize;
-  } else {
-//    std::cout << "Tile for " << Filename(basedirectory, id_) << " not found" << std::endl;
+  }
+  else {
+    LOG_DEBUG("Tile " + file_location + " was not found");
```
dnesbitt61: anything lower than that? This will occur all the time in the builders

gknisely:
:ship:

PR #50, opened 2015-01-21
updated unpaved function.  Is based on surface now.

gknisely:

### DIFF
--- a/src/baldr/directededge.cc
+++ b/src/baldr/directededge.cc
@@ -66,14 +66,7 @@ bool DirectedEdge::destonly() const {
 }
 
 bool DirectedEdge::unpaved() const {
-
-  Surface edge_surface = this->surface();
-  if (edge_surface == Surface::kPavedSmooth ||
-      edge_surface == Surface::kPaved ||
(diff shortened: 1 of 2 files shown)


src/baldr/directededge.cc
```diff
@@ -66,7 +66,14 @@ bool DirectedEdge::destonly() const {
...
-  return false;  //TODO Fix.
+
+  Surface edge_surface = this->surface();
+  if (edge_surface == Surface::kPavedSmooth ||
+      edge_surface == Surface::kPaved ||
+      edge_surface == Surface::kPavedRough)
+    return false;
+
+  return true;
 }
```
dnesbitt61: can't this be: 
return surface < = Surface::kPavedRough

dnesbitt61: I mean:  return (surface() <= Surface::kPavedRough);  
simple! no temp variables or if statements

gknisely:
agree...hold

gknisely:
actually...should be this as you said you wanted cobblestone included as paved.
return (surface() <= Surface::kCompacted);

PR #54, opened 2015-01-26
Gk version

gknisely:
Still need to figure out how to implement in the edgeinfo class as the constructor takes a ptr.  

This version logic can determine when you take a bit(s) from the spare bits and update the structure.  Also, it can determine if you add a member var to the class.  What it can't do, is determine if changed the order of the bitfields in the struct.  

Logic for bitfields is, negate and find the index of the most significant bit to get the "size". Finally, combine it to the seed.

### DIFF
--- a/src/baldr/graphtileheader.cc
+++ b/src/baldr/graphtileheader.cc
@@ -1,4 +1,8 @@
 #include "baldr/graphtileheader.h"
+#include "baldr/nodeinfo.h"
+#include "baldr/directededge.h"
+#include "baldr/edgeinfo.h"
+#include "config.h"
 
 namespace valhalla{
 namespace baldr{
@@ -7,7 +11,17 @@ GraphTileHeader::GraphTileHeader()
     : nodecount_(0),
       directededgecount_(0),
       edgeinfo_offset_(0),
-      textlist_offset_(0) {
+      textlist_offset_(0),
+      internal_version_(0) {
+
+  NodeInfo n;
+  DirectedEdge de;
+  GraphId id;
+
+  internal_version_ = n.hash_value() + de.hash_value() + id.hash_value();
+
+  version_ = PACKAGE_VERSION;
+
 }
 
 size_t GraphTileHeader::nodecount() const {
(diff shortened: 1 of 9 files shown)


kevinkreiser:
i dont understand how the hash_value methods dont change their values depending on what is set in the given objects. some tests would be nice to prove me wrong.

shouldnt those hash methods be static so as to ensure that the fields aren't set (ie instantiate a blank struct) and use it to compute the size of the fields. also please call the methods something other than hash_value so as to discourage their use as a hash. the point of a hash of course is something that is differnt for every valuation of the object. whaat we want is somethign that is the same as long as the  structure of the object is the same.

gknisely:
No you are correct.  It could change if someone set the stuct.  I was only using via a blank struct.  So you think I should make them static and rename?

src/baldr/graphtileheader.cc
```diff
@@ -7,7 +11,17 @@ GraphTileHeader::GraphTileHeader()
     : nodecount_(0),
       directededgecount_(0),
       edgeinfo_offset_(0),
-      textlist_offset_(0) {
+      textlist_offset_(0),
+      internal_version_(0) {
```
kevinkreiser: i would make the internal_version the first thing in the header so that even if the header radically changes we should be abel to read out the version. it shouldnt radically change but hey just in case. same with the other version info, time stamp, version string etc should be right after that. just my opinion

gknisely:
@kevinkreiser @dgearhart @dnesbitt61 
Ok.  Updated.  Again, I don't think this will work for edgeinfo.  We need to look at "poor mans" protobuf and possibly toss this code.  But at least we have some sort of version for now.

gknisely:
Let me know if good.  @dgearhart @kevinkreiser @dnesbitt61

kevinkreiser:
my opinion is its good enough for now yep

dnesbitt61:
I'm not in favor of doing something complex at this time. I think we can
manually track data format changes for the time being. The creation date is
more important at this time.

On Mon, Jan 26, 2015 at 9:54 AM, gknisely <notifications@github.com> wrote:

> @kevinkreiser <https://github.com/kevinkreiser> @dgearhart
> <https://github.com/dgearhart> @dnesbitt61 <https://github.com/dnesbitt61>
> Ok. Updated. Again, I don't think this will work for edgeinfo. We need to
> look at "poor mans" protobuf and possibly toss this code. But at least we
> have some sort of version for now.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/valhalla/baldr/pull/54#issuecomment-71472285>.
>



-- 
David Nesbitt
dave@mapzen.com
443-206-1819

PR #56, opened 2015-01-26
Add internal flag to directed edge.

dnesbitt61:
In preparation for trying to derive this during graph building. Just used a bit in a spare field so should not require rebuilding graph tiles.


dgearhart:
:+1:

PR #63, opened 2015-01-30
constify graphtile and edgeinfo

kevinkreiser:
these are supposed to be readonly classes so i wanted to make sure their interface was true to that.


dnesbitt61:
:shipit:

dgearhart:
:+1:

PR #65, opened 2015-02-02
Change internal storage of lat,lng 

dnesbitt61:
Change to use std::pair of floats but cast toPointLL in read access method.


dgearhart:
:+1:

PR #69, opened 2015-02-03
New exit list logic.

dnesbitt61:
Added GraphId to the GraphTileHeader.
This is a new GetExitSignList method. When it is working the existing exit sign logic (with exits stored on EdgeInfo) will be removed.

### DIFF
--- a/src/baldr/directededge.cc
+++ b/src/baldr/directededge.cc
@@ -26,9 +26,9 @@ GraphId DirectedEdge::endnode() const {
 
 // ------------------  Data offsets and flags for extended data -------------//
 
-// Get the offset to the common edge data.
-uint32_t DirectedEdge::edgedataoffset() const {
-  return dataoffsets_.edgedataoffset;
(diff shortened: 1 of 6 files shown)


src/baldr/directededge.cc
```diff
@@ -24,9 +24,16 @@ GraphId DirectedEdge::endnode() const {
   return endnode_;
 }
 
+// ------------------  Data offsets and flags for extended data -------------//
+
 // Get the offset to the common edge data.
 uint32_t DirectedEdge::edgedataoffset() const {
```
kevinkreiser: i always wondered why this wasnt named edgeinfooffset

dnesbitt61: Good point. Probably a big change but should do it

kevinkreiser: yes a very annoying change that breaks all the builds but might be for the best.

dgearhart:
:+1:

PR #70, opened 2015-02-04
simplify the graphid interface

kevinkreiser:
i was sick of having to call set with all the parameters when all i wanted was to just set the damned id of the element within the tile. i let all of the method accessors so as to minimize diffs (except value because its at the top level of the object). prs for others coming up


dgearhart:
:+1:

gknisely:
:ship:

PR #71, opened 2015-02-04
Move exit signs to be separate entity 

dnesbitt61:
Fixed size so store them after directed edges. Store count rather than offset in GraphTileHeader. Remove exit signs from edgeInfo. Signature of getting exit signs from tile remains the same, but they will be accessed via directed edge index (logic for searching is TBD).

### DIFF
--- a/valhalla/baldr/edgeinfo.h
+++ b/valhalla/baldr/edgeinfo.h
@@ -9,8 +9,6 @@
 #include <valhalla/midgard/pointll.h>
 #include <valhalla/midgard/util.h>
 #include <valhalla/baldr/graphid.h>
-#include <valhalla/baldr/exitsign.h>
-#include <valhalla/baldr/exitsigninfo.h>
 
 using namespace valhalla::midgard;
 
@@ -22,7 +20,6 @@ constexpr size_t kMaxEncodedShapeSize = 16383;
 /**
  * Edge information not required in shortest path algorithm and is
  * common among the 2 directions.
- * @author  David W. Nesbitt
  */
 class EdgeInfo {
  public:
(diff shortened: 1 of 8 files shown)


valhalla/baldr/exitsign.h
```diff
@@ -29,8 +36,10 @@ class ExitSign {
 
  protected:
   // Constructor
-  ExitSign(const ExitSign::Type& type, uint32_t text_offset);
+  ExitSign(const uint32_t idx, const ExitSign::Type& type,
+           const uint32_t text_offset);
 
+  uint32_t edgeindex_;
```
kevinkreiser: i keep wondering if stuff like this should become structs with const members. then we could just get rid of the accessors

valhalla/baldr/exitsign.h
```diff
@@ -21,6 +21,13 @@ class ExitSign {
...
 
+  /**
+   * Get the index of the directed edge this exit sign applies to.
+   * @return  Returns the directed edge index (within the same tile
+   *          as the exit information).
+   */
+  uint32_t edgeindex() const;
+
   // Returns the exit sign type
   const ExitSign::Type& type() const;
```
kevinkreiser: probably dont really need a const reference for the enum since its just an integer

valhalla/baldr/edgeinfo.h
```diff
@@ -83,10 +68,9 @@ class EdgeInfo {
   union PackedItem {
     struct Fields {
 
-      uint32_t name_count                     :4;
-      uint32_t encoded_shape_size             :14;
-      uint32_t exit_sign_count                :4;
-      uint32_t spare                          :10;
+      uint32_t name_count          :4;
+      uint32_t encoded_shape_size  :14;
+      uint32_t spar                :14;
```
dgearhart: spare typo

dgearhart:
:+1:

PR #75, opened 2015-02-06
Add binary search for exit signs.

dnesbitt61:


gknisely:
:+1:

src/baldr/graphtile.cc
```diff
@@ -219,27 +219,43 @@ std::vector<std::string> GraphTile::GetNames(const uint32_t edgeinfo_offset) con
...
-    if (signs_[i].edgeindex() != idx) {
-      continue;
+  // Binary search
+  uint32_t low = 0, high = count-1;
+  uint32_t mid;
+  while (low <= high) {
+    mid = (low + high) / 2;
+    if (signs_[mid].edgeindex() == idx) {
+      // Back up while prior is equal (or at the beginning)
+      while (mid > 0 && signs_[mid-1].edgeindex() == idx) {
```
kevinkreiser: pedant alert (couldnt resist): this technically makes it no longer o(logn) but rather o(n), but in practice this will be better than o(n) basically all the time so who cares

kevinkreiser: you could just stop here and below add in both directions and then the search part would still be logn ;)

PR #77, opened 2015-02-10
avoid allocating memory over  and avoid reallocating containers on expansion

kevinkreiser:
these were popping up when checking hot spots with callgrind


dgearhart:
:+1:

PR #79, opened 2015-02-12
add pathlocation equality method so as to use in a test

kevinkreiser:


src/baldr/pathlocation.cc
```diff
@@ -38,5 +39,22 @@ namespace baldr{
     vertex_ = vertex;
   }
 
+  bool PathLocation::operator==(const PathLocation& other) const {
+    if(node_ != other.node_ || vertex_.ApproximatelyEqual(other.vertex_))
```
dgearhart: was this correct? vertex_.ApproximatelyEqual(other.vertex_) return false?

dgearhart:
:+1:

PR #80, opened 2015-02-13
Add support for simple turn restrictions.

dnesbitt61:
A flag on the directed edge indicates the start of a simple turn restriction. A bit mask in the turn restriction record (indexed by the directed edge index) indicates the indexes of directed edges at the end node which are restricted. Turn restriction records are fixed size and stored as an array (ordered by the from directed edge index). A binary search is used to access - here there is only one TR record possible for a directed edge (unlike the sign lookup) which simplifies things.


gknisely:
:ship:

PR #85, opened 2015-02-18
Rework simple turn restrictions

dnesbitt61:
Simple turn restrictions are stored in the directed edge, not a separate piece of data in the graph tile.


kevinkreiser:
:ship:

gknisely:
:ship:

PR #87, opened 2015-02-18
Add access methods for most of the new data members.

dnesbitt61:


gknisely:
:ship:

kevinkreiser:
would it be crazy if we just made this into a struct, removed most of the methods and made public the various structs to be accessed directly? IMHO it feels like there is not much point to having these methods as they mostly dont do anything but return a value. i guess im saying these objects dont really seem like classes, they seem like simple structs to me.

dgearhart:
:+1:

PR #88, opened 2015-02-20
Use 8 bits for turn restriction mask 

dnesbitt61:
- set a max so we can see if it is ever exceeded in tile building.


gknisely:
:ship:

dgearhart:
:+1:

PR #89, opened 2015-02-20
Fix size of type_

dnesbitt61:
Add IntersectionType, local_driveable (count). Fix spare so size of type... was incorrect (was 1 bit too large!).


dgearhart:
:+1:

PR #90, opened 2015-02-20
Add test to make sure size of NodeInfo is correct.

dnesbitt61:
Tested this by adding 1 bit to one of the "spare" parts of a struct so it exceeded the 32 bit size.
This caused a test failure - which is what we want.


dgearhart:
:+1:

PR #96, opened 2015-02-26
Fixes for Asia

dnesbitt61:
Validate and fix lat,lng inputs - throw exception if lng is outside range from -360 to 360 degrees or lat is outside -90 to 90 degrees. Fix longitude to lie within -180 to 180 degrees.

Increase max edge info offsets. Believe there are a few very dense areas in Asia where we don't drop enough edges going from local to arterial.


kevinkreiser:
:boat:

gknisely:
:ship:

PR #97, opened 2015-02-27
added payment.

gknisely:


dnesbitt61:
Need to have indications of what each bit means. 8 bits is a lot - can we distill into a smaller set?

gknisely:
Is a mask.
// Payment constants. Bit constants.
constexpr uint8_t kCoins  = 1; // Coins
constexpr uint8_t kNotes  = 2; // Bills
constexpr uint8_t kETC    = 4; // Electronic Toll Collector

dnesbitt61:
Added earlier I guess. So you only need 3 bits?

gknisely:
I guess but could be all three...should we add credit cards and other (bitcoin, paypal, etc)

PR #100, opened 2015-03-02
Added turn type

dgearhart:
Updated odin and tests to follow


kevinkreiser:
looks good to me, but i wonder why we even need a class for it? maybe it could just be a namespace level method? unless in the future we need to compare two turns to eachother or something?

dgearhart:
I was thinking about possible future functionality

kevinkreiser:
sweet sounds good

PR #104, opened 2015-03-03
Needed 4 bits for opposing local index. 

dnesbitt61:
Any value > max transitions 8 will not be able to look up edge transition information, but we need a value > 8
so a valid value is not used in cases where > 8 edges occur at a node.


dgearhart:
:+1:

PR #108, opened 2015-03-06
qaudruple the encoded shape size max, up the max density so that at leas...

kevinkreiser:
...t no NA nodes will complain


dgearhart:
:shipit:

dnesbitt61:
:shipit

PR #110, opened 2015-03-06
Updates for edge transitions

dnesbitt61:
NodeInfo size increased to 32bytes: added name continuity, headings, and driveability for up to 8 edges (using local edge index to reference).
DirectedEdge size reduce to 40 bytes: added edge_to_left and edge_to_right flags but removed name continuity and spare 64bit word reserved for edge transitions (not needed due to additions to NodeInfo).


dgearhart:
:shipit:

PR #111, opened 2015-03-09
Change local driveable to local_edge_count 

dnesbitt61:
(for how many edge heading/name consistency are stored)


dgearhart:
:shipit:

PR #115, opened 2015-03-12
Add is_shortcut flag. 

dnesbitt61:
Only kMaxShortcutsFromNode can have a mask and supersede an edge, but more than this # of shortcuts can be stored.


dgearhart:
:shipit:

PR #116, opened 2015-03-13
moved StreetName and StreetNames from odin to baldr

dgearhart:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,6 +44,8 @@ test/graphid
 test/tilehierarchy
 test/graphtile
 test/nodeinfo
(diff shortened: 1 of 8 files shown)


dnesbitt61:
Will this work outside US?

dgearhart:
there is a TODO to implement based on locale

dnesbitt61:
:shipit

kevinkreiser:
is there anyway we could mark this one as the US one and then use it as default until we have a real default?

dgearhart:
I will work on refactor

PR #119, opened 2015-03-20
Admins

gknisely:

Admin updates for baldr.

### DIFF
--- a/valhalla/baldr/admininfo.h
+++ b/valhalla/baldr/admininfo.h
@@ -59,19 +59,22 @@ class AdminInfo {
   // Packed items: counts for names and level
   union PackedItem {
     struct Fields {
-      uint32_t name_count          :8;
-      uint32_t timezone            :6;
-      uint32_t dst                 :1;
-      uint32_t spare               :17;
-
+      uint64_t start_dst            :24; //MMDDHHMM
+      uint64_t end_dst              :24; //MMDDHHMM
+      uint64_t name_count           :8;  //name count
+      uint64_t parent_admin         :6;  //parent admin index
+      uint64_t spare                :2;
     } fields;
-    uint32_t value;
+    uint64_t value;
   };
 
  protected:
 
   // Where we keep the statistics about how large the vectors below are
   PackedItem* item_;
+
+  // ISO 3166-2 code.  Index into textlist
+  uint32_t iso_code_index_;
 
   // List of name indexes
   uint32_t* name_offset_list_;
--- a/valhalla/baldr/admininfo.h
+++ b/valhalla/baldr/admininfo.h
@@ -72,10 +72,10 @@ class AdminInfo {
   PackedItem* item_;
 
   // DST start date and time.
-  std::string start_dst_;
+  char start_dst_[9];
 
   // DST end date and time.
-  std::string end_dst_;
+  char end_dst_[9];
 
   // ISO 3166-2 code.  Index into textlist
   uint32_t iso_code_index_;
--- a/valhalla/baldr/admininfo.h
+++ b/valhalla/baldr/admininfo.h
@@ -1,0 +1,90 @@
+#ifndef VALHALLA_BALDR_ADMININFO_H_
+#define VALHALLA_BALDR_ADMININFO_H_
+
+#include <vector>
+#include <string>
+#include <ostream>
+#include <iostream>
+
+#include <valhalla/midgard/pointll.h>
+#include <valhalla/midgard/util.h>
+#include <valhalla/baldr/graphid.h>
+
+using namespace valhalla::midgard;
+
+namespace valhalla {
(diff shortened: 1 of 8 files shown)


valhalla/baldr/admininfo.h
```diff
@@ -0,0 +1,90 @@
...
+    struct Fields {
+      uint32_t name_count          :8;
+      uint32_t timezone            :6;
+      uint32_t dst                 :1;
+      uint32_t spare               :17;
+
+    } fields;
+    uint32_t value;
+  };
+
```
dnesbitt61: Admin info does NOT contain timezone or DST flag. It should contain the start/end dates for DST. Each node has to contain a timezone and dst flag (not at admin level). See: https://docs.google.com/a/mapzen.com/document/d/1ZVJ3Ptbl0BxIvg-7Xh_tY_LY88TISt7di0LkKSSnZWg/edit

valhalla/baldr/admininfo.h
```diff
@@ -0,0 +1,90 @@
...
+ protected:
+
+  // Where we keep the statistics about how large the vectors below are
+  PackedItem* item_;
+
+  // List of name indexes
+  uint32_t* name_offset_list_;
+
+  // The list of names within the tile
+  const char* names_list_;
```
dgearhart: Do we have two text lists? We are not able to add names to existing text list?

dnesbitt61:
Did we ditch the idea of storing ISO country codes and abbreviations? If we think we might want to add in the future then it should be built in now.

valhalla/baldr/admininfo.h
dnesbitt61: I don't want to have separate names lists. When we add transit we can just append names and have one place that all names exist. We can de-dup incoming names against each other, but perhaps not go through the pain of de-dup against existing names

valhalla/baldr/admininfo.h
gknisely: Not sure about that at admin level....for now I will leave with namecount

valhalla/baldr/admininfo.h
gknisely: old code...removing.  good catch

gknisely:
Names are saved as <name>, en:<name>, etc

dnesbitt61:
Yes, but there will be one ISO code. If we are going to store it then add it to the base Admin structure rather than have a name lookup

valhalla/baldr/admininfo.h
dnesbitt61: ?? I meant remove timezone and DST. Add fields for start and end DST date

dgearhart:
I thought we were using the ISO 3166-1 alpha-2 for each country

valhalla/baldr/admininfo.h
gknisely: sorry not old code....the admins are reusing the textlist

dnesbitt61:
Goal is to add the fields we think we will want/need. Populate as many as we can now and leave others TODO. But that way we don't change the graph tile format each time we want to populate a field. So I think a 2 char field should be added to the base AdminInfo structure, for the ISO code of the country .

gknisely:
So what do you want in PackedItem?  You guys are going back and forth.  
Count
Start/End dates for DST 
and what else?

dnesbitt61:
Looks to me like admin info should contain:
Start DST
End DST
ISO code
name count
name offsets (variable size)

valhalla/baldr/admininfo.h
```diff
@@ -0,0 +1,93 @@
...
+    struct Fields {
+      uint64_t start_dst            :24; //MMDDHHMM
+      uint64_t end_dst              :24; //MMDDHHMM
+      uint64_t name_count           :8;  //name count
+      uint64_t parent_admin         :6;  //parent admin index
+      uint64_t spare                :2;
+    } fields;
+    uint64_t value;
+  };
+
```
dnesbitt61: What is the form of start and end dst and how are they converted to usable values? Wouldn't it be better to have these as separate data members outside PackedItem? Where are the access methods to read al the values (I only see name count and nameoffset(index).

gknisely:
03080200 begins
11010200 ends
I was going add access functions next pull request for balr as I need to populate parent_admin and iso offset sooner rather than later.

dnesbitt61:
Is that a standard way to represent a date? Seesm it would require building a string and then changing it to the numeric representation?

gknisely:
I am trying to save space.

dnesbitt61:
There are only a couple admin areas per tile, so the savings is a handfl of bytes. Judt seems to be a non-standard (and not trivial) conversion

gknisely:
updated.

dgearhart:
:+1:

PR #121, opened 2015-03-31
Add transit stops and routes.

dnesbitt61:
Not used for path finding but for trip path building.

### DIFF
--- a/src/baldr/transitroute.cc
+++ b/src/baldr/transitroute.cc
@@ -1,0 +1,64 @@
+#include "baldr/transitroute.h"
+#include <boost/functional/hash.hpp>
+#include <cmath>
+
+namespace valhalla {
+namespace baldr {
+
+// Default constructor
+TransitRoute::TransitRoute()
+    : routeid_(0),
+      agencyid_(0),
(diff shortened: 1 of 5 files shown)


src/baldr/transitroute.cc
```diff
@@ -0,0 +1,64 @@
+#include "baldr/transitroute.h"
+#include <boost/functional/hash.hpp>
```
kevinkreiser: why the unused headers?

dnesbitt61: For Greg's hashing scheme for testing internal version...

src/baldr/transitroute.cc
```diff
@@ -0,0 +1,64 @@
...
+#include <cmath>
+
+namespace valhalla {
+namespace baldr {
+
+// Default constructor
+TransitRoute::TransitRoute()
```
kevinkreiser: in place of these constructors we can indeed just instantiate these objects like `TransitRoute some_transit_route{};` and it will have the same effect

dgearhart:
:shipit:

PR #122, opened 2015-03-31
Add Message Passing Tools For Service Robustness

kevinkreiser:
this adds some tools around zmq that will allow us to setup our service in a distributed manner. there is a test which also simulates a service that computes prime numbers


dnesbitt61:
:ship:

dgearhart:
:shipit:

PR #123, opened 2015-03-31
Transit departures

dnesbitt61:
Fixed size. Still needs some date/time definitions, trying to make them small enough to compress easily yet still allow easy lookup.


dgearhart:
:+1:

PR #124, opened 2015-04-01
More transit structure additions.

dnesbitt61:
Moved a couple things into graphconstants. Nothing is actually in the graph tile or graphtile header yet, just putting in some base structures.


dgearhart:
:shipit:

PR #125, opened 2015-04-02
Refactored StreetNames to have base and US specific class

dgearhart:


gknisely:
Are we going to have a class for every country or is this temporary?

dgearhart:
We will have a default and one for any country that needs it. I will add a factory for that based on country code

gknisely:
:ship:

PR #127, opened 2015-04-06
Admin updates

gknisely:
@kevinkreiser @dgearhart @dnesbitt61  

Admins are now fixed size.  Country and state names are local.  Saved as indexes to text_offset list.


valhalla/baldr/admin.h
```diff
@@ -0,0 +1,79 @@
...
+  uint32_t country_offset() const;
+
+ protected:
+  // Constructor
+  Admin(const uint32_t country_offset, const uint32_t state_offset,
+        const std::string& country_iso, const std::string& state_iso,
+        const std::string& start_dst, const std::string& end_dst);
+
+  // country ISO3166-1
+  char country_iso_[kCountryIso];
```
dnesbitt61: byte alignment. Since this is an odd number of bytes can you find out the size? Arrange objects and add spare so it is explicit.

dnesbitt61:
:shipit:

PR #131, opened 2015-04-09
Added phone, url, and heading

dgearhart:

### DIFF
--- a/src/baldr/location.cc
+++ b/src/baldr/location.cc
@@ -45,12 +45,17 @@ namespace baldr{
 
     //grab some address info
     auto part = parts.begin() + 2;
-    for(auto address : { &l.name_, &l.street_, &l.city_, &l.state_, &l.zip_, &l.country_ }) {
+    for(auto address : { &l.name_, &l.street_, &l.city_, &l.state_, &l.zip_, &l.country_, &l.phone_, &l.url_ }) {
       if(part == parts.end())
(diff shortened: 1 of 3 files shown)


gknisely:
:sheep:

src/baldr/location.cc
```diff
@@ -45,12 +45,17 @@ namespace baldr{
 
     //grab some address info
     auto part = parts.begin() + 2;
-    for(auto address : { &l.name_, &l.street_, &l.city_, &l.state_, &l.zip_, &l.country_ }) {
+    for(auto address : { &l.name_, &l.street_, &l.city_, &l.state_, &l.zip_, &l.country_, &l.phone_, &l.url_ }) {
```
kevinkreiser: i thought the plan was that, since we dont actually have any of these extra attributes in the data, that we wouldnt even store them. that we would only just pull out the ones that we use/need and the others would be just a pass through to be echoed back on output?

dgearhart: my understanding is - we would just pass  along so informationis  packaged together

kevinkreiser: we have no idea what the format will be though right? this makes it so the format is specific and only handles a finit number of attributes. my opinion is that we should parse out only what we want to worry about and anything else gets forwarded through (thereby not limiting what else gets forwarded through)

PR #133, opened 2015-04-11
updated to use unique_ptr.

gknisely:

### DIFF
--- a/src/baldr/graphtile.cc
+++ b/src/baldr/graphtile.cc
@@ -251,22 +251,22 @@ std::vector<std::string> GraphTile::GetNames(const uint32_t edgeinfo_offset) con
 }
 
 // Get the admininfo at the specified index.
-std::unique_ptr<const AdminInfo> GraphTile::admininfo(const size_t idx) const {
+const AdminInfo GraphTile::admininfo(const size_t idx) const {
   if (idx < header_->admincount()) {
     const Admin admin = admins_[idx];
 
-    return std::unique_ptr<AdminInfo>(new AdminInfo(textlist_ + admin.country_offset(),
-                                                    textlist_ + admin.state_offset(),
-                                                    admin.country_iso(), admin.state_iso(),
-                                                    admin.start_dst(), admin.end_dst()));
+    return AdminInfo(textlist_ + admin.country_offset(),
+                     textlist_ + admin.state_offset(),
+                     admin.country_iso(), admin.state_iso(),
+                     admin.start_dst(), admin.end_dst());
   }
   throw std::runtime_error("GraphTile AdminInfo index out of bounds");
 }
 
 // Get the admin at the specified index.
-const Admin* GraphTile::admin(const size_t idx) const {
+const Admin GraphTile::admin(const size_t idx) const {
   if (idx < header_->admincount()) {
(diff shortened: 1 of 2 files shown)


src/baldr/graphtile.cc
```diff
@@ -251,14 +251,22 @@ std::vector<std::string> GraphTile::GetNames(const uint32_t edgeinfo_offset) con
...
-const AdminInfo* GraphTile::admininfo(const size_t idx) const {
+std::unique_ptr<const AdminInfo> GraphTile::admininfo(const size_t idx) const {
   if (idx < header_->admincount()) {
     const Admin admin = admins_[idx];
 
-    return new AdminInfo(textlist_ + admin.country_offset(),
-                         textlist_ + admin.state_offset(),
-                         admin.country_iso(), admin.state_iso(),
-                         admin.start_dst(), admin.end_dst());
+    return std::unique_ptr<AdminInfo>(new AdminInfo(textlist_ + admin.country_offset(),
```
dgearhart: as i mentioned yesterday - i added a make_unique in midgard util. I think we should use the make_* methods instead of the direct use of 'new'. It uses pefect-forwards so your parameters will just pass through to constructor. 

return make_unique<AdminInfo>(textlist_ + admin.country_offset(), textlist_ + admin.state_offset(), admin.country_iso(), admin.state_iso(),admin.start_dst(), admin.end_dst());

when including midgard util.h using midgard

dgearhart: return make_unique<AdminInfo>(textlist_ + admin.country_offset(), textlist_ + admin.state_offset(), admin.country_iso(), admin.state_iso(),admin.start_dst(), admin.end_dst());

dgearhart: "return make_unique<AdminInfo>(textlist_ + admin.country_offset(), textlist_ + admin.state_offset(), admin.country_iso(), admin.state_iso(),admin.start_dst(), admin.end_dst());"

gknisely: If this is the case then why hasn't the code been updated?  edgeinfo(const size_t offset) const uses the same logic.

dgearhart: make_uniques was just added recently - so yes, we should update as we can

dgearhart:
:shipit:

src/baldr/graphtile.cc
dnesbitt61: Why return a pointer in the first place?

gknisely:
@dgearhart @dnesbitt61 @kevinkreiser returning obj instead of ptr for Admins now.

src/baldr/graphtile.cc
```diff
@@ -251,14 +251,22 @@ std::vector<std::string> GraphTile::GetNames(const uint32_t edgeinfo_offset) con
 }
 
 // Get the admininfo at the specified index.
-const AdminInfo* GraphTile::admininfo(const size_t idx) const {
+const AdminInfo GraphTile::admininfo(const size_t idx) const {
   if (idx < header_->admincount()) {
     const Admin admin = admins_[idx];
```
dnesbitt61: Use a reference rather than copying

PR #136, opened 2015-04-14
Added FromPtree and associated tests

dgearhart:

### DIFF
--- a/src/baldr/location.cc
+++ b/src/baldr/location.cc
@@ -2,9 +2,11 @@
 #include <stdexcept>
 #include <boost/algorithm/string/split.hpp>
 #include <boost/algorithm/string/classification.hpp>
+#include <boost/property_tree/json_parser.hpp>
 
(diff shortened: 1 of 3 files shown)


dnesbitt61:
:ship:

src/baldr/location.cc
```diff
@@ -12,10 +14,61 @@ namespace baldr{
   Location::Location(const midgard::PointLL& latlng, const StopType& stoptype):latlng_(latlng), stoptype_(stoptype){
   }
 
-  Location Location::FromJson(const std::string& json){
-    throw std::runtime_error("Location serialization from json is not yet implemented");
+Location Location::FromPtree(const boost::property_tree::ptree& pt) {
```
kevinkreiser: any reason why you didnt want the bits about taking a string and turning into a ptree here as well?

dgearhart: we could add if needed

PR #137, opened 2015-04-14
SRTM - load a tile and access height at a lat,lng. Simple test program.

dnesbitt61:


gknisely:
:ship:

kevinkreiser:
are we going to use it from mjolnir through here? is it going to be its own service? im wondering if it should be its own thing instead of in baldr

dnesbitt61:
Yes. Yes. Want to create a service, but also want mjolnir to be able to access SRTM tiles directly when building graph tiles. The stuff I added to baldr is the simplest, single tile reading with bilinear filtering of the 4 nearest posts. Will need to add a layer in mjolnir to cache SRTM tiles and manage access to the lowest level.

kevinkreiser:
yeah i was just wondering if the separate repo could have both the service and the library that could do all of these things

dnesbitt61:
Forgot to merge this - added some transit methods to access records by Id.

dgearhart:
:shipit:

PR #140, opened 2015-04-16
default to break for location type

kevinkreiser:
updated tests, removed explicit testing of `FromPtree` since the test was just doing exactly what `FromJson` does, ie. make a ptree and then just call `FromPtree`


gknisely:
:shaved_ice:

PR #141, opened 2015-04-16
good catch @dgearhart, if unintelligable stop type use break. added test...

kevinkreiser:
... for this as well


dgearhart:
:shipit:

PR #150, opened 2015-05-07
Updates to GetNextDeparture based on rules for calendar date exceptions.

dnesbitt61:
Calendar additions are added with start_date == end_date and only subtractions
remain in the TransitCalendar records.


dgearhart:
:shipit:

PR #152, opened 2015-05-07
Add convenience method "is_transit" 

dnesbitt61:
Check if the node is a transit stop when returning the stop_id (returns 0 if not a transit stop).


gknisely:
:train:

PR #154, opened 2015-05-10
move json to baldr from tyr

kevinkreiser:
need it at various stages the move was long overdue. would have rather had it in midgard but it depends on boost::variant and i figured it wasnt worth making midgard have boost dependencies (and it certainly wasnt worth reimplementing variant!)

added some serializing of path location object so that loki can pass those onto thor


dgearhart:
:shipit:

PR #156, opened 2015-05-12
Methods needed to change destination edges at a node

dnesbitt61:
They need to be the incoming edges. Add a way to clear the existing edges and then we can add back the opposing edge Ids in thor when needed.


dgearhart:
:shipit:

PR #158, opened 2015-05-13
Add methods to get a map of populated tiles and a connectivity map.

dnesbitt61:
This will allow testing origin and destination to see if a path of populated tiles connects them. We can very quickly reject paths between locations that have no possible connectivity. To handle long ferries we can add "empty" tiles for where the ferry crosses but no end nodes exist.

Also added a method to return the tile Id given the filename (no need to actually open/read the file).


kevinkreiser:
you'll want to add libboost-filesystem1.54-dev to the scripts/depedencies.sh

PR #160, opened 2015-05-14
Transit

gknisely:
added GetName and GetTransitDeparture

### DIFF
--- a/src/baldr/graphtile.cc
+++ b/src/baldr/graphtile.cc
@@ -489,14 +489,18 @@ const TransitDeparture* GraphTile::GetTransitDeparture(const uint32_t edgeid,con
 
   if (found) {
     // Back up while prior is equal (or at the beginning)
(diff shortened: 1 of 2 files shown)


src/baldr/graphtile.cc
```diff
@@ -445,6 +455,57 @@ const TransitDeparture* GraphTile::GetNextDeparture(const uint32_t edgeid,
...
+        departures_[mid-1].edgeid() == edgeid) {
+      mid--;
+    }
+
+    while (departures_[mid].tripid() != tripid && mid < count) {
+      mid++;
+    }
+    if (departures_[mid].tripid() == tripid)
+      return &departures_[mid];
+  }
```
dnesbitt61: While backing up can check for == tripid and return. If not found then start at saved position and work forward?

dgearhart:
:shipit:

PR #162, opened 2015-05-19
update connectivity map to be singleton

kevinkreiser:
moved the connectivity map into a singleton in graphgreader. it gets configured once and used everywhere. also changed the reverse mapping of file name string to graphid (was previously hardcoded to be level two with .25 degrees or whatever we have in our configs)

with test yay


dnesbitt61:
:shipit

PR #163, opened 2015-05-20
sort geojson connectivity map by largest regions first

kevinkreiser:
last connectivity map pull request for now i promise (excluding the one to turn it on in data production chef land of course)

### DIFF
--- a/src/baldr/connectivity_map.cc
+++ b/src/baldr/connectivity_map.cc
@@ -70,10 +70,11 @@ namespace {
     });
   }
 
-  std::string to_feature_collection(const std::unordered_map<size_t, std::list<PointLL> >& regions) {
+  template <class T>
+  std::string to_feature_collection(const std::unordered_map<size_t, std::list<PointLL> >& regions, const std::multimap<size_t, size_t, T>& arities) {
     auto features = json::array({});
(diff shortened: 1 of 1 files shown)


src/baldr/connectivity_map.cc
```diff
@@ -70,10 +70,11 @@ namespace {
     });
   }
 
-  std::string to_feature_collection(const std::unordered_map<size_t, std::list<PointLL> >& regions) {
+  template <class T>
+  std::string to_feature_collection(const std::unordered_map<size_t, std::list<PointLL> >& regions, const std::multimap<size_t, size_t, T>& arities) {
     auto features = json::array({});
-    for(const auto& region : regions)
-      features->emplace_back(to_feature(region));
+    for(const auto& arity : arities)
```
kevinkreiser: write the features with the largest arities first. so that the largest colored regions appear first in the geojson. this should make it easier to have stable colors for the largest regions when rendering. the largest regions shouldnt change much and should pretty much be in the same order. so if we generate colors randomly with a constant seed, the largest features will likely get the same colors with the smaller regions jockying for the last places in the list and seeing more color changes

gknisely:
:ship:

PR #166, opened 2015-05-22
Fix TransitDeparture and move to using line Id rather than edge Id

dnesbitt61:
Store a unique line Id in directed edges and use that rather than the directed edge Id.
Fix a bug with TransitDeparture (used a union rather than struct for ScheduleDates).
Update transit use values to be at the upper end of the range.


kevinkreiser:
ship

PR #175, opened 2015-06-16
Time goodies

gknisely:
First step in TimeZone logic.  2 functions added:  Get the iso date and time from a DOW mask and time. Get the iso date and time from the current time.  Test added as well.  

I am thinking that the functions will have to be re-factored to use a tz offset from UTC and not the tz_string.  We will see.....anyhow, like I said first steps.


dnesbitt61:
:shipit:

PR #179, opened 2015-06-29
Clang compatibility fixes

fsaric:
I managed to compile and run valhalla (pathtest and pbfgraphbuilder) on MacOS 10.10 w/clang, but there are some minor fixes needed. For example `hash_combine` can't take a reference to `uint64_t` and `nullptr_t` is defined in std namespace. Example error:

```
src/baldr/directededge.cc:328:3: error: no matching function for call to 'hash_combine'
  boost::hash_combine(seed, ffs(de.dataoffsets_.edgeinfo_offset+1)-1);
  ^~~~~~~~~~~~~~~~~~~
/opt/local/include/boost/functional/hash/hash.hpp:310:17: note: candidate function [with T = int] not viable: no known conversion from
      'uint64_t' (aka 'unsigned long long') to 'std::size_t &' (aka 'unsigned long &') for 1st argument
    inline void hash_combine(std::size_t& seed, T const& v)
```


kevinkreiser:
This whole thing brings to mind another issue. We will be storing this number in the final tile. The issue with size_t, and why we never store that, is that it can potentially vary from system to system (but the tiles need to be read on any system). This is why they use a type with specified bit width. I guess what i'm saying is, if you make tiles on a machine with 32bit prec for size_t for example, and a 64bit machine goes to read them, you run a pretty decent risk of the tiles not being compatible.

Basically we need to re-implement the versioning stuff because its not robust to different environments.

That said, this was already the case, the thing is our build system is happily turning uint64_t into size_t and back without complaint. So your fix is still valid although unfortunately just as prone to the issue.

fsaric:
The trouble is that boost::hash_combine works only on a `size_t &` and it returns different results on 32bit and 64bit systems, so it should probably be replaced.

kevinkreiser:
yep exactly. that is what i meant with the bit about "re-implement the versioning stuff"

PR #187, opened 2015-08-20
rename elevation to grade and set default

kevinkreiser:


dnesbitt61:
:shipit

kevinkreiser:
will ship once dependencies match and are pr'd

PR #189, opened 2015-08-21
added support for transitland's date format.  Updated tests.

gknisely:
added support for transitland's date format.
added tests as needed.


dgearhart:
:shipit:

PR #194, opened 2015-09-09
add timezone data to a header

kevinkreiser:
this will create a header `valhalla/baldr/date_time_zonespec.h` from the data found at https://raw.githubusercontent.com/boostorg/date_time/master/data/date_time_zonespec.csv

the header will look like 

    unsigned char date_time_zonespec_csv[] = {
      0x22, 0x49, 0x44, 0x22, 0x2c, 0x22, 0x53, 0x54, 0x44, 0x20, 0x41, 0x42,
      0x42, 0x52, 0x22, 0x2c, 0x22, 0x53, 0x54, 0x44, 0x20, 0x4e, 0x41, 0x4d,
    ...
      0x22, 0x2c, 0x22, 0x22, 0x2c, 0x22, 0x22, 0x2c, 0x22, 0x22, 0x2c, 0x22,
      0x2b, 0x30, 0x30, 0x3a, 0x30, 0x30, 0x3a, 0x30, 0x30, 0x22, 0x0a
    };
    unsigned int date_time_zonespec_csv_len = 35855;

then wherever @gknisely needs to load this data all he must do is loop over this binary data, write it to file, and then send that temp file to boost to load up the timezone stuff. this makes it so we dont have to guess where it is on the system.

i suggest we call wahtever boost function with this tempfile and store the resulting object in a static initializer (singleton) in an anonymous namespace and have whatever namespace level function just return results from that.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,6 +35,8 @@ srtmtest
 *.lo
 .libs/
 libvalhalla_baldr.la
+date_time_zonespec.csv
+valhalla/baldr/date_time_zonespec.h
 
 # tests
 test-suite.log
--- a/Makefile.am
+++ b/Makefile.am
@@ -35,6 +35,13 @@ clean-gcno:
 clean-local: clean-coverage clean-gcno
 endif
 
+#timezone info from boost
+valhalla/baldr/date_time_zonespec.h:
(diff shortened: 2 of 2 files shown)


gknisely:
:watch:

PR #195, opened 2015-09-10
Tz

gknisely:
Using the results from the previous PR:  https://github.com/valhalla/baldr/pull/194#event-405063898
to load the boost date_time_zonespec.csv file so that we can get tz and dst info.

### DIFF
--- a/src/baldr/datetime.cc
+++ b/src/baldr/datetime.cc
@@ -1,23 +1,50 @@
 #include <iostream>
 #include <sstream>
 
-#include "boost/date_time/posix_time/posix_time.hpp"
-#include "boost/date_time/gregorian/gregorian.hpp"
+#include <boost/date_time/posix_time/posix_time.hpp>
+#include <boost/date_time/gregorian/gregorian.hpp>
 #include <boost/range/algorithm/remove_if.hpp>
 #include <boost/algorithm/string/classification.hpp>
 #include <boost/date_time/posix_time/posix_time.hpp>
 #include <boost/date_time/local_time/local_time.hpp>
 #include <boost/date_time/local_time/local_time_io.hpp>
 #include <boost/algorithm/string.hpp>
+#include <boost/filesystem/operations.hpp>
 
 #include <valhalla/baldr/datetime.h>
 #include <valhalla/baldr/graphconstants.h>
+#include <valhalla/baldr/date_time_zonespec.h>
+#include <fstream>
 
 namespace valhalla {
 namespace baldr {
 namespace DateTime {
 
-boost::gregorian::date pivot_date_ = boost::gregorian::from_undelimited_string(kPivotDate);
+namespace
+{
+static boost::local_time::tz_database* tz_db_ = nullptr;
+static boost::gregorian::date pivot_date_ = boost::gregorian::from_undelimited_string(kPivotDate);
+
+static boost::local_time::tz_database* get_tz_db() {
+  if(!tz_db_) {
+
+    boost::system::error_code error;
+    auto tmp_path = boost::filesystem::temp_directory_path(error);
+    std::string file;
+//try to save the file in the /tmp dir
+    if (error.value() == boost::system::errc::success) {
+      file = tmp_path.c_str();
+      file += "/";
+    }
+    file += "valhalla_tz.csv";
+    tz_db_ = new boost::local_time::tz_database();
+    std::ofstream outfile (file,std::ofstream::binary);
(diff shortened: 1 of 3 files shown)


src/baldr/datetime.cc
```diff
@@ -1,23 +1,50 @@
...
+    std::string file;
+//try to save the file in the /tmp dir
+    if (error.value() == boost::system::errc::success) {
+      file = tmp_path.c_str();
+      file += "/";
+    }
+    file += "valhalla_tz.csv";
+    tz_db_ = new boost::local_time::tz_database();
+    std::ofstream outfile (file,std::ofstream::binary);
+    outfile.write(reinterpret_cast<char*>(&date_time_zonespec_csv),date_time_zonespec_csv_len);
```
kevinkreiser: you should close the file before reading it. and you should also delete it when done

src/baldr/datetime.cc
```diff
@@ -1,23 +1,50 @@
...
 namespace baldr {
 namespace DateTime {
 
-boost::gregorian::date pivot_date_ = boost::gregorian::from_undelimited_string(kPivotDate);
+namespace
+{
+static boost::local_time::tz_database* tz_db_ = nullptr;
+static boost::gregorian::date pivot_date_ = boost::gregorian::from_undelimited_string(kPivotDate);
+
+static boost::local_time::tz_database* get_tz_db() {
```
kevinkreiser: none of the stuff in the anonymous namespace should be static

src/baldr/datetime.cc
```diff
@@ -1,23 +1,50 @@
...
+    boost::system::error_code error;
+    auto tmp_path = boost::filesystem::temp_directory_path(error);
+    std::string file;
+//try to save the file in the /tmp dir
+    if (error.value() == boost::system::errc::success) {
+      file = tmp_path.c_str();
+      file += "/";
+    }
+    file += "valhalla_tz.csv";
+    tz_db_ = new boost::local_time::tz_database();
```
kevinkreiser: why is this a pointer?

src/baldr/datetime.cc
```diff
@@ -1,23 +1,50 @@
...
+#include <fstream>
 
 namespace valhalla {
 namespace baldr {
 namespace DateTime {
 
-boost::gregorian::date pivot_date_ = boost::gregorian::from_undelimited_string(kPivotDate);
+namespace
+{
+static boost::local_time::tz_database* tz_db_ = nullptr;
```
kevinkreiser: delete this

src/baldr/datetime.cc
```diff
@@ -1,23 +1,50 @@
...
+static boost::gregorian::date pivot_date_ = boost::gregorian::from_undelimited_string(kPivotDate);
```
kevinkreiser: should be constexpr not static

src/baldr/datetime.cc
```diff
@@ -134,6 +161,10 @@ std::string iso_date_time(const uint8_t dow_mask, const std::string& time,
       break;
   }
 
+  boost::local_time::tz_database* tz_db = get_tz_db();
```
kevinkreiser: make this static and not a pointer

gknisely:
K hold

src/baldr/datetime.cc
```diff
@@ -1,23 +1,50 @@
...
+//try to save the file in the /tmp dir
+    if (error.value() == boost::system::errc::success) {
+      file = tmp_path.c_str();
+      file += "/";
+    }
+    file += "valhalla_tz.csv";
+    tz_db_ = new boost::local_time::tz_database();
+    std::ofstream outfile (file,std::ofstream::binary);
+    outfile.write(reinterpret_cast<char*>(&date_time_zonespec_csv),date_time_zonespec_csv_len);
+    tz_db_->load_from_file(file);
```
kevinkreiser: looks like it has a method load_from_stream, so instead of making a file just use the char array into a stream

src/baldr/datetime.cc
```diff
@@ -134,6 +161,10 @@ std::string iso_date_time(const uint8_t dow_mask, const std::string& time,
       break;
   }
 
+  boost::local_time::tz_database* tz_db = get_tz_db();
+  if (!tz_db)
+    throw std::runtime_error("Unable to load boost's date_time_zonespec file.");
```
kevinkreiser: i imagine this might happen above in the line tz_db_.load_from_file()

PR #196, opened 2015-09-10
Traversability

gknisely:
Driveability -----> Traversability

### DIFF
--- a/valhalla/baldr/nodeinfo.h
+++ b/valhalla/baldr/nodeinfo.h
@@ -95,12 +95,12 @@ class NodeInfo {
   bool dst() const;
 
   /**
-   * Get the auto traversability of the local directed edge given a local
+   * Get the driveability of the local directed edge given a local
    * edge index.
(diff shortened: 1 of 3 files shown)


dnesbitt61:
I don't like renaming local_driveability). The method name would need to be: local_auto_traversability which just seems long and confusing to me! I don't mind changing the enum name and explain in the comment it is based on the mode of travel.

gknisely:
@dgearhart loves  local_auto_traversability i am sure

valhalla/baldr/nodeinfo.h
```diff
@@ -221,7 +221,7 @@ class NodeInfo {
 
   // Node type
   struct NodeTypeInfo {
-    uint32_t local_driveability : 16; // Driveability for local edges (up to
+    uint32_t auto_traversability : 16; // Traversability for local edges (up to
```
dgearhart: should we keep the local prefix?

dgearhart:
:shipit:

PR #197, opened 2015-09-15
Timezone

gknisely:


valhalla/baldr/nodeinfo.h
```diff
@@ -14,7 +14,7 @@ namespace baldr {
 constexpr uint32_t kMaxTileEdgeCount    = 4194303;  // 2^22 directed edges
 constexpr uint32_t kMaxEdgesPerNode     = 127;      // Maximum edges per node
 constexpr uint32_t kMaxAdminsPerTile    = 63;       // Maximum Admins per tile
-constexpr uint32_t kMaxTimeZonesPerTile = 63;       // Maximum TimeZones per tile
+constexpr uint32_t kMaxTimeZonesPerTile = 450;      // Maximum TimeZones index
```
dnesbitt61: 9 bits allows up to 511 timezone entries

valhalla/baldr/nodeinfo.h
```diff
@@ -212,10 +212,9 @@ class NodeInfo {
 
   // Administrative information
   struct NodeAdmin {
+    uint16_t timezone     : 9; // Time zone
     uint16_t admin_index  : 6; // Index into this tile's list of admin data
-    uint16_t timezone     : 6; // Time zone
     uint16_t dst          : 1; // Is Daylight Saving Time used?
-    uint16_t spare        : 3;
   };
   NodeAdmin admin_;
```
dnesbitt61: This breaks the service - keep admin_index first then the timezone field can have 9 bits without breaking existing code. Also is the DST flag needed? Isn't that part of the timezone record?

valhalla/baldr/nodeinfo.h
gknisely: Correct, but the db is at around 430.  But I can change to 511

valhalla/baldr/nodeinfo.h
gknisely: I don't think DST is needed.  I was going to discuss on Thurs, but I can update now

valhalla/baldr/nodeinfo.h
dnesbitt61: Yes, change to what the max that can be specified within the field (511). Then in the builder make sure this max is not exceeded (log error, etc.) when setting the field.

gknisely:
We are going to push this logic out with transit updates as it breaks the server version test.

PR #201, opened 2015-10-02
self serializers for nodes and edges

kevinkreiser:
this is useful for locate results. the main reason we might want this (at least why i want it) is so that we can far more easily debug issues. knowing how something is attributed is a massive help. ive purposely left a bit of information from both nodes and edges out (not knowing if its useful or visually possibly to determine whats what). anyway we can expand on it at any time.
a


dgearhart:
:shipit:

dnesbitt61:
:ship:

PR #202, opened 2015-10-16
Increase max speed.

dnesbitt61:
Now set to ~85MPH. Several places in US and internationally have this speed limit. Have not seen any higher. This has slight impact to A* / performance for driving routes.


dgearhart:
:+1:

PR #204, opened 2015-12-07
Transit and Tile Format Changes

kevinkreiser:
Lots of data format changes to support additional transit data (but also removed a couple of no longer needed structures). Also no longer have "builders" in mjolnir so there are set methods for many classes.


dnesbitt61:
:shipit;

PR #205, opened 2015-12-07
is_iso_local function

gknisely:
Added is_iso_local function to validate the date_time that the user sends to loki.  
Added tests.

Date_time_value string will be validated in loki.


dnesbitt61:
:shipit:

PR #206, opened 2015-12-09
added flag to check if the date that was inputed is before the tile_creation date.

gknisely:
We need this because date can be less than tile_creation date and in MMcosting we were doing the calculation of day = date - tile_creation_date_;  And day is unsigned.  

Thor updates coming as well.


dgearhart:
:shipit:

PR #218, opened 2016-01-15
make subdivisions in tile configurable

kevinkreiser:
the number of subdivisions for a single tile for edge binning needs to come from config


dnesbitt61:
:shipit:

kevinkreiser:
still a ship?

PR #220, opened 2016-01-20
make nominclature match across projects

kevinkreiser:
since we've been calling this edge binning it seemed we should be consistent throughout the projects. this will make mjolnir fail so id only merge it once i have the edge binning branch of mjolnir pr'd and ready


dnesbitt61:
:shipit:

PR #221, opened 2016-01-25
Remove bus and rail node types for now.

dnesbitt61:
fixes issue #200


gknisely:
will this break anything?

dgearhart:
I was wondering the same thing

dnesbitt61:
These 2 types are not present in code anywhere.  Since the enum has explicit values assigned the enum values do not change (there are now just 2 empty spots). I changed the is_transit method as well to now just check for the MultiUseStop type.

dgearhart:
:shipit:

gknisely:
:train:

PR #223, opened 2016-02-02
Validate tile header density value.

dnesbitt61:
Move kMaxDensity to graphconstants.h so it can be used for both node and tile density.


gknisely:
:+1:

PR #224, opened 2016-02-17
don't download date_time_zonespec.csv anymore.

gknisely:


dnesbitt61:
Can you document where the data comes from so it is easy to update periodically?

gknisely:
@dnesbitt61 updated the readme   Let me know if it looks good

dnesbitt61:
:shipit:

PR #225, opened 2016-02-17
Also set reverseaccess to all

dnesbitt61:
(this is used for transition edges and this change makes it so access can be checked on opposing edges to make sure they match (forward against reverse on opposing edge and vice-versa).


kevinkreiser:
:rocket:

PR #231, opened 2016-03-15
routeid --> route_type 

gknisely:
routeid not needed.  using this space to set the route_type now.  
Also, checked in a missing header updated from date_time.


dnesbitt61:
:shipit:

PR #232, opened 2016-03-15
GetOpposingEdgeId - return invalid GraphId for transit lines

dnesbitt61:
Return an invalid GraphId when getting the opposing edge of a transit line. These do not have proper opposing edges assigned yet.


dgearhart:
:ship:

PR #233, opened 2016-03-18
add convenient bin by index for loki

kevinkreiser:


dgearhart:
any tests to validate your pair? otherwise :ship:

kevinkreiser:
testing was all manual in mjolnir, but thinking about it some more there is a way to test it with a mocked up derived tile class, so i'll give that a whirl, thanks!

kevinkreiser:
now with tests, thanks @dgearhart

PR #234, opened 2016-03-22
TransitDeparture Rework

dnesbitt61:
Add a structure to store unique schedule validity periods. 
Replace TransitTransfer (unused) with the new TransitSchedule structure within tiles.
Separate transit schedule information from transit departures and index into the new TransitSchedule records. 
Compress TransitDeparture to get down to 20 bytes per departure (was 32).
Check max. values and throw exceptions when exceeded.


gknisely:
:ship:

PR #239, opened 2016-04-12
Fix JSON output for access and bike network.

dnesbitt61:
These are bitwise & operations


kevinkreiser:
my mistake, sorry and thanks for fixing :rocket:

PR #244, opened 2016-05-02
remove requirement of gcc4.9 or greater by using identical boost regex

kevinkreiser:
ok so i know we dont really want to add more boost dependencies unless absolutely necessary. its much nicer to make use of stdlib (which slowly incorporates boost) because we dont have to get that it comes with the compiler. but there is a subtle issue here.

`trusty` the version of ubuntu that is default on amazon and a lot of people are still using comes with `gcc-4.8`. this version of the compiler did not support regex (although it did make available the declarations interface etc). the first version of the compiler supporting it properly was the next version `gcc-4.9`. this is fine, we can and indeed do get gcc-4.9 and run with it. the problem is two fold.

first having a dependency on ppas means that any user (through `apt-get`) of `valhalla` will have those same dependencies. this is annoying but it cant be avoided due to `valhalla` using `prime_server`. i would argue though that we do want to minimize this. we could also fork `prime_server` and build it with `valhalla`, some other issues there as well.

second, and this is the major one, c++11 was considered experimental before `gcc-5.0` this means that their implementations were subject to change. specifically between 4.8 and 4.9 the sizes of several c++11 objects changed and its pretty much impossible to get properly compatible libraries built with 4.8 linked into appplications built 4.9.

appologies :frowning:


dgearhart:
:shipit:

kevinkreiser:
i forgot to mention that all of this can go away once we get onto a system that has 5.0 or greater. so if we move to xenial for example we'l be all good :smile:

PR #246, opened 2016-05-09
need to reencode to visible ascii

kevinkreiser:
when we had visible ascii shape and it was compressed in the `edgeinfo`, then the `loki::locate` action could just use it directly. but now its in a more compact format. so to return it we must decode it and reencode it.


gknisely:
:ship:

PR #249, opened 2016-05-25
Send in tile directory to GetTileId (rather than TileHierarchy).

dnesbitt61:
Should allow this to be used by transit tiles.


dgearhart:
:ship:

PR #250, opened 2016-05-25
Update to allow connectivity on transit level.

dnesbitt61:


dgearhart:
:shipit:

src/baldr/connectivity_map.cc
```diff
@@ -87,6 +87,9 @@ namespace {
 namespace valhalla {
   namespace baldr {
     connectivity_map_t::connectivity_map_t(const TileHierarchy& tile_hierarchy):tile_hierarchy(tile_hierarchy) {
+      // Set the transit level
+      transit_level = tile_hierarchy.levels().rbegin()->second.level + 1;
+
       // Populate a map for each level of the tiles that exist
       for (const auto& tile_level : tile_hierarchy.levels()) {
```
kevinkreiser: cant you just add `3` to the levels for loop here and avoid duplicated this code and having to track transit separately?

PR #252, opened 2016-06-02
connectivity map as polygons instead of point clouds

kevinkreiser:
before the geojson output was points and was not great for efficiency nor display. now its polygons which is far easier to use.

caveat, i havent yet implemented the inner rings of the polygons and therefore you may see two different connectivity regions (specifically smaller ones inside of big ones) that touch eachother. This is because one of them is inside of a missing inner ring of another one. i'll get that in another pr.

ive also styled the geojson so that it displays nicely on geojson.io. once these are up on s3 we can use geojson.ios api to display the latest ones directly from a github.io page. i'll hack that together at some point..

### DIFF
--- a/src/baldr/connectivity_map.cc
+++ b/src/baldr/connectivity_map.cc
@@ -10,6 +10,8 @@
 #include <sstream>
 #include <unordered_set>
 
+#include <valhalla/midgard/logging.h>
+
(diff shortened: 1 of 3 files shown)


dnesbitt61:
:shipit:

PR #253, opened 2016-06-08
Access updates

gknisely:
updated uses for country admin logic.


valhalla/baldr/graphconstants.h
```diff
@@ -205,7 +205,10 @@ enum class Use : uint8_t {
 
   // Pedestrian specific uses
   kFootway = 25,
-  kSteps   = 26,           // Stairs
+  kPedestrian = 26,
+  kPath = 27,
+  kSteps = 28,             // Stairs
```
kevinkreiser: any reason for changing kSteps to 28? you could have kPedestrian be 28 and keep steps the same avoid confusion? maybe it doesnt matter at all though :smile:

gknisely: yes to keep the foot paths together....that is all

kevinkreiser: looks like it doesnt really matter though in the end, just used for a switch case in mjolnir so all good

kevinkreiser:
:ship:

valhalla/baldr/graphconstants.h
dnesbitt61: Do not change values of existing uses if you can help it. That would mean that server and data producer would have to be deployed together (steps is used in costing and costing filters)

dnesbitt61: Also - are pedestrian and path basically footway after all logic is complete? Or would this require several additional if statements in costing to determine if the edge should be favored for pedestrian routes?

gknisely:
updated

PR #254, opened 2016-06-10
refactor for fuzzy connectivity map and loki islands multiple results

kevinkreiser:
this refactors connectivity map lookups out of graph reader because the way in which loki will use it is changing.

it also refactors pathlocation so that loki can more easily have multiple matches (for islands but also for fuzzy matching and other improvements related to search scoring)

### DIFF
--- a/src/baldr/connectivity_map.cc
+++ b/src/baldr/connectivity_map.cc
@@ -249,6 +249,29 @@ namespace valhalla {
       return color->second;
     }
 
+    std::unordered_set<size_t> connectivity_map_t::get_colors(uint32_t hierarchy_level,
+      const baldr::PathLocation& location, float radius) const {
+
+      std::unordered_set<size_t> result;
+      auto level = colors.find(hierarchy_level);
+      if(level == colors.cend())
+        return result;
+      const auto& tiles = tile_hierarchy.levels().find(hierarchy_level)->second.tiles;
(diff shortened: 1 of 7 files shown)


gknisely:
ship when circle good

ptpt:
Is a `PathLocation` is a "mouse click" location with multiple edge locations correlated nearby?

kevinkreiser:
yep, it can be a mouse click or a geocode or a reverse geocode or anything like that. and yes it can be multiple edge locations correlated nearby but i havent implemented that part yet. these are the changes needed to allow for it. for more info on why we want it, see here: valhalla/loki#127

ptpt:
I see. Thanks for clarifying it.

In meili a candidate is a thin wrapper of PathLocation but a candidate
actually is defined as one edge location. I probably need to some
changes there to avoid confusion.

kevinkreiser:
Oh no its fine to use it has a holder for only one location, its just been extended to allow more. If meili ended up calling loki it will be able to control loki to only ever correlate to one spot. you should be safe to use it without modification :smile:

kevinkreiser:
alternatively we could add a bit to PathLocation that says how many different spots its correlated to. this would allow you to throw or warn or whatever makes sense if ever its not the case that its only 1. again that should never happen though

PR #257, opened 2016-06-27
fix for 32bit systems

kevinkreiser:
so it turns out this was a bad way to test. since json dictionaries are unordered it doesnt make sense to check that the output is exactly the same. so ive cheated and made the test take the two outputs, sort the lines of output and check that each line is the same. the ordering of things in the unordered data structures on 32bit machines was just slightly different.


dgearhart:
:shipit:

PR #264, opened 2016-08-05
this allows Location to be hashed

kevinkreiser:
this is useful for loki so that if you are doing matrix and have duplicates they wont be searched more than once. it uses the intuition that the most likely different thing between a set of locations is the latlon and uses the equality operator to differentiate the rest.

this also remove const from the returned object of `GetNames` in `EdgeInfo`. It basically never makes sense to return a const value type. Or my feeble brain cant figure out when it would!


kdiluca:
:+1:  thank you!

gknisely:
:ship:

PR #265, opened 2016-08-05
Schedule fix

gknisely:
Old logic would add schedule to the data and start it at the tile date and not the true start date.  This caused unnecessary transfers.  New tests added.

Amtrak transfer issue:

screenshot from 2016-08-05 09 58 30


kevinkreiser:
:trolleybus:

PR #268, opened 2016-08-10
Add a flag for a directed edge that ends at a dead-end node.

dnesbitt61:
Needed to allow U-turns at dead-ends


kevinkreiser:
:rocket:

PR #271, opened 2016-08-11
Set maximum values for trip Id and headsign offset.

dnesbitt61:
Will allow us to work with bit fields to allow higher elapsed time values.


gknisely:
:ship:

dgearhart:
:shipit:

PR #272, opened 2016-08-15
tz offset reverted

gknisely:
added offset back and tz to the end....thor will remove tz and send b…ack via protocol buffer.


dgearhart:
:shipit:

PR #275, opened 2016-09-07
strenghting tile null checks

kevinkreiser:
this only fixes one of many possible. we should possibly also try hardening: `AreEdgesConnected` and `GetEdgeDensity` against missing data


dnesbitt61:
:shipit:

PR #276, opened 2016-09-08
Moved from thor

dnesbitt61:
Removed the dependency on sif EdgeLabels. Now accepts a function to get the sortcost from a specified label index within the empty_overflow method.


dgearhart:
:shipit: when compiled

PR #277, opened 2016-09-09
added onestop maps.

gknisely:
used to include or exclude transit lines.

### DIFF
--- a/src/baldr/graphtile.cc
+++ b/src/baldr/graphtile.cc
@@ -146,6 +146,44 @@ LOG_INFO("Departures: " + std::to_string(header_->departurecount()) +
     textlist_ = graphtile_.get() + header_->textlist_offset();
     textlist_size_ = filesize - header_->textlist_offset();
 
+    //if this tile is transit then we need to save off the pair<tileid,lineid> lookup via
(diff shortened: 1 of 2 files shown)


src/baldr/graphtile.cc
```diff
@@ -146,6 +146,44 @@ LOG_INFO("Departures: " + std::to_string(header_->departurecount()) +
...
+    //if this tile is transit then we need to save off the pair<tileid,lineid> lookup via
+    //onestop_ids.  This will be used for including or excluding transit lines for transit
+    //routes.  We save 2 maps because operators contain all of their route's tile_line pairs
+    //and it is used to include or exclude the operator as a whole.
+    if (graphid.level() == 3) {
+
+      for (uint32_t i = 0; i < header_->directededgecount(); i++) {
+        if (directededges_[i].lineid() != 0) {
+
+          const auto& dep = GetTransitDeparture(directededges_[i].lineid());
```
kevinkreiser: shame on the compiler, i wish it would error for these. i think this should be `const auto*` for the sake of not exploding peoples brains ;)

gknisely:
@kevinkreiser thanks for the help with the removal of the binary search for departures

dnesbitt61:
:shipit:

PR #280, opened 2016-09-14
add a score to pathlocation

kevinkreiser:
can use this for fuzzy matching and allows use in meili


dnesbitt61:
:shipit:

kevinkreiser:
fixes #278

PR #286, opened 2016-09-21
load an extract of tiles via a memory mapped tar

kevinkreiser:
this first time you start your server the singleton will take about 20 seconds to index the tar. on subsequent starts it will take a fraction of a second.


dnesbitt61:
:shipit:

src/baldr/graphreader.cc
```diff
@@ -52,33 +77,39 @@ const GraphTile* GraphReader::GetGraphTile(const GraphId& graphid) {
   //TODO: clear the cache automatically once we become overcommitted by a certain amount
 
   // Check if the level/tileid combination is in the cache
-  auto cached = cache_.find(graphid.Tile_Base());
+  auto base = graphid.Tile_Base();
```
kevinkreiser: instead of calling `Tile_Base` over and over just cache it

PR #287, opened 2016-09-21
Add output for wheelchair. Validate set access.

dnesbitt61:
Check access modes to make sure they do not exceed the max (kAllAccess).


src/baldr/directededge.cc
```diff
@@ -26,6 +26,7 @@ json::MapPtr access_json(uint32_t access) {
     {"pedestrian", static_cast<bool>(access & kPedestrianAccess)},
     {"taxi", static_cast<bool>(access & kTaxiAccess)},
     {"truck", static_cast<bool>(access & kTruckAccess)},
+    {"wheelchair", static_cast<bool>(access && kWheelchairAccess)}
```
dgearhart: should this be bitwise and

dgearhart:
:ship:

PR #288, opened 2016-09-21
warn if there are corrupt blocks in the tar

kevinkreiser:


src/baldr/graphreader.cc
```diff
@@ -30,6 +30,9 @@ struct GraphReader::tile_extract_t : public midgard::tar {
         LOG_WARN("Could not load tile extract");
         return;
       }
+      //some corrupt stuff
+      if(corrupt_blocks)
+        LOG_WARN(std::to_string(corrupt_blocks) + " corrupt blocks in tile extract");
```
dnesbitt61: Perhaps check if contents is not empty before saying Tile extract loaded.

kevinkreiser: sounds good!

dnesbitt61:
:shipit:

PR #290, opened 2016-09-22
Add method to associate onestop Ids.

dnesbitt61:
This is called from the Initialize method if the tile is a transit tile. This was not in prior MMAP logic, now this works for both mmap and file I/O.

### DIFF
--- a/src/baldr/graphtile.cc
+++ b/src/baldr/graphtile.cc
@@ -84,47 +84,7 @@ GraphTile::GraphTile(const TileHierarchy& hierarchy, const GraphId& graphid)
     file.close();
 
     // Set pointers to internal data structures
-    Initialize(graphtile_.get(), filesize);
-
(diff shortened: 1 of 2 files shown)


gknisely:
:+1:

src/baldr/graphtile.cc
```diff
@@ -206,6 +158,51 @@ LOG_INFO("Departures: " + std::to_string(header_->departurecount()) +
 
   // Set the size to indicate success
   size_ = tile_size;
+
+  // Associate one stop Ids for transit tiles
+  if (graphid.level() == 3) {
```
kevinkreiser: might be good to get this via `TileHeirarchy.levels()` call?

kevinkreiser:
comment isnt a show stopper just a suggestion, :rocket:

PR #291, opened 2016-09-22
make binary search faster, add wheelchair and bike to departure lookup

kevinkreiser:
instead of spooling back to the beginning to get the first matching record ive slightly modified the 4 binary searches to find the first matching record. ive also simplified the logic after each one for the various other criteria.

in addition to that when asking for the next departure you can also specify that you need one that supports bike and/or wheelchair

### DIFF
--- a/valhalla/baldr/graphtile.h
+++ b/valhalla/baldr/graphtile.h
@@ -206,8 +206,8 @@ class GraphTile {
                                            const uint32_t day,
                                            const uint32_t dow,
                                            bool  date_before_tile,
-                                           bool wheelchair = false,
-                                           bool bicycle = false) const;
+                                           bool wheelchair,
+                                           bool bicycle) const;
 
   /**
    * Get the departure given the directed edge Id and tripid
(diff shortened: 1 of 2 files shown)


valhalla/baldr/graphtile.h
```diff
@@ -196,14 +196,18 @@ class GraphTile {
...
    * @return  Returns a pointer to the transit departure information.
    *          Returns nullptr if no departures are found.
    */
   const TransitDeparture* GetNextDeparture(const uint32_t lineid,
                                            const uint32_t current_time,
                                            const uint32_t day,
                                            const uint32_t dow,
-                                           bool  date_before_tile) const;
+                                           bool  date_before_tile,
+                                           bool wheelchair = false,
```
gknisely: no defaults needed.

kevinkreiser: i only did it so that i wouldnt break the build. do you want me to remove it?

gknisely: I will always set it, but up to you....if you leave make both true.

dnesbitt61:
:shipit:

PR #293, opened 2016-09-23
Make compilation outside of the project directory possible

TeXitoi:
For context, see kevinkreiser/prime_server#50

baldr generate files. I've make the choice to put generated files in the genfiles directory.

What do you think of that?


kevinkreiser:
its a fine idea but it may break valhalla/packaging. so we'll need to work around that.

also instead of making blank targets in the else statement could you just define clean-local differently depending on whether or not test coverage was enabled. like this:

```diff
31,34d30
< else
< .PHONY: clean-coverage
< clean-coverage:
< endif
39a36,40
> clean-local: clean-coverage clean-gcno clean-genfiles
> else
> clean-local: clean-genfiles
> endif
> 
42,43d42
< 
< clean-local: clean-coverage clean-gcno clean-genfiles
```

TeXitoi:
OK, looking at valhalla/packaging makes me understand that touching all these build processes may not be a good idea :wink:

How do you manage different compiling flavor on your dev environment (debug, release, release with debug info...)?

kevinkreiser:
i just do it by passing the various flags to configure similar to this: https://stackoverflow.com/questions/4553735/gnu-autotools-debug-release-targets#4680578

TeXitoi:
But that's what I was trying to do with these PR: being able to compile outside of the source dir. Thus I suppose that you have a clone by compile type or you `make distclean && CXXFLAGS='-g -O2' ./configure && make` ?

kevinkreiser:
oh im sorry i thought you were only asking how do we do release vs debug. i have not yet ran into the need to compile it from outside of the source dir. when we release it we do it one of two ways:

from source - https://github.com/valhalla/tools#building-and-running-valhalla
from ppa - https://github.com/valhalla/packaging/blob/master/package.sh

PR #295, opened 2016-09-23
graphreader is the arbitor of tiles

kevinkreiser:
so connectivity map was broken because it rolled its own t hing for looking at what tiles we have. this is the first step to making graphreader the sole controller of knowing where tiles are (for reading at least). ive added a comment to tilehierarchy for the rest of the work but it has a lot of side effects so i climbed back out of that rabbit hole for now.

### DIFF
--- a/src/baldr/connectivity_map.cc
+++ b/src/baldr/connectivity_map.cc
@@ -1,6 +1,7 @@
 #include "baldr/connectivity_map.h"
 #include "baldr/json.h"
 #include "baldr/graphtile.h"
+#include "baldr/graphreader.h"
 
 #include <valhalla/midgard/pointll.h>
 #include <boost/filesystem.hpp>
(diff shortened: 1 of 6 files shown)


src/baldr/connectivity_map.cc
```diff
@@ -207,35 +208,26 @@ namespace {
 
 namespace valhalla {
   namespace baldr {
-    connectivity_map_t::connectivity_map_t(const TileHierarchy& tile_hierarchy):tile_hierarchy(tile_hierarchy) {
-      // Set the transit level
+    connectivity_map_t::connectivity_map_t(const boost::property_tree::ptree& pt)
+      :tile_hierarchy(pt.get<std::string>("tile_dir")) {
+      // See what kind of tiles we are dealing with here by getting a graphreader
+      GraphReader reader(pt);
```
kevinkreiser: i think graphreader should eventually encapsulate connectivity map, the refactor was too much for me to do right now though

dnesbitt61:
:shipit:

PR #298, opened 2016-09-29
Cache EdgeInfo in `GraphTile::edgeinfo()`

### LINKED ISSUE #297: Performance of loki::Search with lots of entries
Our 1x192 request use a lot of time in `GraphTile::edgeinfos()` and `EdgeInfos::shape()`. `EdgeInfos::shape()` is cached, but `EdgeInfos` is not cached in `GraphTile`.

As we search 192 points, we basically re-generate `EdgeInfos` for each of these 192 points. Thus, caching `GraphTile::edgeinfos()` might improve drastically the performances in our usecase.

I can cache this info using the technique used in `GraphReader::GetGraphTile()`. Are you OK with that? Do you see any reason why that's a bad idea or it will not work? Do you have any other idea to fix this problem?

Thanks in advance.

TeXitoi:
It's working here, improving from 1.46s to 1.27s.

It implies a couple of modification all around valhalla, but nothing evil (changing the type of `edgeinfo()` and `GraphTile` not copyable).

fix #297


kevinkreiser:
this looks good to me and as you've shown has improved performance. i want to let @dnesbitt61 comment as well but I like it as it models the rest of our extra caching on graph tiles.

TeXitoi:
data are our 191 locations from our test request

before:
```
2016/09/30 08:53:19.174953 [INFO] Succeeded Searches on Uncached Tiles
2016/09/30 08:53:19.175021 [INFO] --------------------------------
2016/09/30 08:53:19.175029 [INFO] Total: 191
2016/09/30 08:53:19.175037 [INFO] Fastest: 49.050304,2.074745 @ 5ms
2016/09/30 08:53:19.175042 [INFO] Slowest: 49.036606,2.080227 @ 19ms
2016/09/30 08:53:19.175046 [INFO] Median: 7.539267ms
2016/09/30 08:53:19.175050 [INFO] Standard Deviation: 1.540816ms
2016/09/30 08:53:19.175054 [INFO] Faster Than 1 Standard Deviation: 8
2016/09/30 08:53:19.175057 [INFO] Slower Than 1 Standard Deviation: 11
2016/09/30 08:53:19.175060 [INFO] --------------------------------


2016/09/30 08:53:19.175078 [INFO] Failed Searches on Uncached Tiles
2016/09/30 08:53:19.175081 [INFO] --------------------------------
2016/09/30 08:53:19.175084 [INFO] No results
2016/09/30 08:53:19.175087 [INFO] --------------------------------


2016/09/30 08:53:19.175106 [INFO] Succeeded Searches on Cached Tiles
2016/09/30 08:53:19.175109 [INFO] --------------------------------
2016/09/30 08:53:19.175112 [INFO] Total: 191
2016/09/30 08:53:19.175116 [INFO] Fastest: 49.054451,2.092744 @ 1ms
2016/09/30 08:53:19.175120 [INFO] Slowest: 49.050007,2.081322 @ 6ms
2016/09/30 08:53:19.175123 [INFO] Median: 3.670157ms
2016/09/30 08:53:19.175127 [INFO] Standard Deviation: 0.766511ms
2016/09/30 08:53:19.175130 [INFO] Faster Than 1 Standard Deviation: 25
2016/09/30 08:53:19.175133 [INFO] Slower Than 1 Standard Deviation: 4
2016/09/30 08:53:19.175135 [INFO] --------------------------------


2016/09/30 08:53:19.175147 [INFO] Failed Searches on Cached Tiles
2016/09/30 08:53:19.175150 [INFO] --------------------------------
2016/09/30 08:53:19.175153 [INFO] No results
2016/09/30 08:53:19.175155 [INFO] --------------------------------
```

after:
```
2016/09/30 09:04:56.056735 [INFO] Succeeded Searches on Uncached Tiles
2016/09/30 09:04:56.056807 [INFO] --------------------------------
2016/09/30 09:04:56.056815 [INFO] Total: 191
2016/09/30 09:04:56.056824 [INFO] Fastest: 49.053196,2.094746 @ 5ms
2016/09/30 09:04:56.056829 [INFO] Slowest: 49.036568,2.080131 @ 17ms
2016/09/30 09:04:56.056833 [INFO] Median: 9.340314ms
2016/09/30 09:04:56.056836 [INFO] Standard Deviation: 1.802964ms
2016/09/30 09:04:56.056840 [INFO] Faster Than 1 Standard Deviation: 20
2016/09/30 09:04:56.056843 [INFO] Slower Than 1 Standard Deviation: 14
2016/09/30 09:04:56.056846 [INFO] --------------------------------


2016/09/30 09:04:56.056863 [INFO] Failed Searches on Uncached Tiles
2016/09/30 09:04:56.056866 [INFO] --------------------------------
2016/09/30 09:04:56.056870 [INFO] No results
2016/09/30 09:04:56.056872 [INFO] --------------------------------


2016/09/30 09:04:56.056892 [INFO] Succeeded Searches on Cached Tiles
2016/09/30 09:04:56.056894 [INFO] --------------------------------
2016/09/30 09:04:56.056898 [INFO] Total: 191
2016/09/30 09:04:56.056902 [INFO] Fastest: 49.056198,2.086392 @ 0ms
2016/09/30 09:04:56.056906 [INFO] Slowest: 49.050007,2.081322 @ 5ms
2016/09/30 09:04:56.056910 [INFO] Median: 2.068063ms
2016/09/30 09:04:56.056913 [INFO] Standard Deviation: 0.686882ms
2016/09/30 09:04:56.056916 [INFO] Faster Than 1 Standard Deviation: 31
2016/09/30 09:04:56.056919 [INFO] Slower Than 1 Standard Deviation: 40
2016/09/30 09:04:56.056922 [INFO] --------------------------------


2016/09/30 09:04:56.056934 [INFO] Failed Searches on Cached Tiles
2016/09/30 09:04:56.056937 [INFO] --------------------------------
2016/09/30 09:04:56.056940 [INFO] No results
2016/09/30 09:04:56.056943 [INFO] --------------------------------
```

Summary:
 * without cache: 7.539267ms -> 9.340314ms (+24%)
 * with cache: 3.670157ms -> 2.068063ms (-44%)

So a non-negligible loose without cache, and a big win with cache.

kevinkreiser:
this might make the case for two methods of access. @dnesbitt61 was mentioning in places like thor where edgeinfo is needed but very sparsely it might not be very useful. in other words what if we changed renamed the method in this pr to something like `GraphTile::edgeinfo_cache(size_t edgeinfo_offset)` and kept the original method as is. then we give the option to loki and meili to use the cache since it benefits them but for thor to use the previous method it since it wont frequently access the same shapes (lots of cache misses).

src/baldr/graphtile.cc
```diff
@@ -360,8 +360,15 @@ GraphId GraphTile::GetOpposingEdgeId(const DirectedEdge* edge) const {
 }
 
 // Get a pointer to edge info.
-std::unique_ptr<const EdgeInfo> GraphTile::edgeinfo(const size_t offset) const {
-  return std::unique_ptr<EdgeInfo>(new EdgeInfo(edgeinfo_ + offset, textlist_, textlist_size_));
+const EdgeInfo* GraphTile::edgeinfo(const size_t offset) const {
+  auto edgeinfo_it = edgeinfo_cache_.find(offset);
```
kevinkreiser: i wonder if cold cache slow down is because of reallocation. can you add to the GraphTile constructor something like `edgeinfo_cache_.reserve(header_->directededgecount);` and see if it brings that loss down to basically 0 (id  hope)?

kevinkreiser: so if this does indeed bring the cold cache back to normal timings and we make the other change with the two interfaces to get at edgeinfos, then we have only one more problem to solve. reserving value type EdgeInfo can take a lot of memory per tile. So we could change the edgeinfo_cache_ to store smart pointers instead. this would mean that if thor doesnt use the cache it wont waste that much memory for reservation

kevinkreiser: another option for the optional caching could just be an argument to the function `bool cache=false` and then update the internal logic.

TeXitoi: The reserve has no impact. I think the overhead is in the search in the unordered_map.

I don't understand your idea with smart pointers. Is it just for `edgeinfo()` and `edgeinfo_cache()` to have the same signature?

TeXitoi: funny: the reserve has an impact, but in the cached case... Maybe reserving allow to better distribution or something like that.

TeXitoi:
Bench with reserve:
```
2016/09/30 14:09:05.469879 [INFO] Succeeded Searches on Uncached Tiles
2016/09/30 14:09:05.469958 [INFO] --------------------------------
2016/09/30 14:09:05.469965 [INFO] Total: 191
2016/09/30 14:09:05.469974 [INFO] Fastest: 49.052719,2.095035 @ 6ms
2016/09/30 14:09:05.469979 [INFO] Slowest: 49.046165,2.093357 @ 19ms
2016/09/30 14:09:05.469983 [INFO] Median: 9.507853ms
2016/09/30 14:09:05.469986 [INFO] Standard Deviation: 1.783781ms
2016/09/30 14:09:05.469989 [INFO] Faster Than 1 Standard Deviation: 21
2016/09/30 14:09:05.469992 [INFO] Slower Than 1 Standard Deviation: 22
2016/09/30 14:09:05.469994 [INFO] --------------------------------


2016/09/30 14:09:05.470011 [INFO] Failed Searches on Uncached Tiles
2016/09/30 14:09:05.470014 [INFO] --------------------------------
2016/09/30 14:09:05.470016 [INFO] No results
2016/09/30 14:09:05.470019 [INFO] --------------------------------


2016/09/30 14:09:05.470037 [INFO] Succeeded Searches on Cached Tiles
2016/09/30 14:09:05.470039 [INFO] --------------------------------
2016/09/30 14:09:05.470042 [INFO] Total: 191
2016/09/30 14:09:05.470046 [INFO] Fastest: 49.051617,2.099416 @ 0ms
2016/09/30 14:09:05.470050 [INFO] Slowest: 49.026489,2.098492 @ 3ms
2016/09/30 14:09:05.470053 [INFO] Median: 1.575916ms
2016/09/30 14:09:05.470056 [INFO] Standard Deviation: 0.581788ms
2016/09/30 14:09:05.470059 [INFO] Faster Than 1 Standard Deviation: 3
2016/09/30 14:09:05.470062 [INFO] Slower Than 1 Standard Deviation: 6
2016/09/30 14:09:05.470065 [INFO] --------------------------------


2016/09/30 14:09:05.470076 [INFO] Failed Searches on Cached Tiles
2016/09/30 14:09:05.470078 [INFO] --------------------------------
2016/09/30 14:09:05.470081 [INFO] No results
2016/09/30 14:09:05.470083 [INFO] --------------------------------
```

src/baldr/graphtile.cc
kevinkreiser: the idea with smart pointers is that they will take less space to reserve than reserving EdgeInfo objects since they just hold a pointer and a reference count, 2 64bit items. Reserving EdgeInfos will cost more because they have 6 64bit members (mostly pointers) and one vector (who knows how big that is). So to keep the cache as pointers will be much cheaper than keeping it as value types.

TeXitoi: > References and pointers to either key or data stored in the container are only invalidated by erasing that element, even when the corresponding iterator is invalidated.

Thus, I suspect your `std::shared_ptr` optimization will not work. If it works,  `std::unique_ptr` is more appropriate and takes less space.

ptpt: what about caching edge infos in a vector of unique pointers?

TeXitoi: that's worst: using a 
 * reserved `std::vector`: 43.879581ms/1.989529ms
 * `std::vector`: 36.141361ms/1.900000ms
 * `std::deque`: 19.167539ms/1.513089ms

kevinkreiser: yep using smart pointers as the value type of the hash map doesnt cost less. i just tested it. they must be implementing the buckets as linked lists or something. sorry for the bad suggestion

src/baldr/graphtile.cc
```diff
@@ -360,8 +363,16 @@ GraphId GraphTile::GetOpposingEdgeId(const DirectedEdge* edge) const {
 }
 
 // Get a pointer to edge info.
-std::unique_ptr<const EdgeInfo> GraphTile::edgeinfo(const size_t offset) const {
-  return std::unique_ptr<EdgeInfo>(new EdgeInfo(edgeinfo_ + offset, textlist_, textlist_size_));
+const EdgeInfo* GraphTile::edgeinfo(const size_t offset) const {
```
kevinkreiser: so can you just add another parameter here something like `bool cache` and default it to false in the declaration. then just in loki (and in meili if it makes sense) call this function with `cache=true`. then all you n eed to do is add a branch to either use the cache look up or just make an object on the fly? we would still need to wrap it in a unique_ptr or something so as to keep the deinitialization handled automatically. if thats too hairy maybe we can at least make a clear edge info cache function so thor can clear it after forming the trip path (which accesses the edgeinfo)

TeXitoi: You mean something like that (I can't think of a version without `std::shared_ptr` that doesn't implies dirty hacks based on dynamic unique_ptr custom deleter):
```c++
std::shared_ptr<const EdgeInfo> GraphTile::edgeinfo(const size_t offset, const bool with_cache = false) const;
```

Personally, I'd prefer using another function, like that:
```c++
std::unique_ptr<const EdgeInfo> GraphTile::edgeinfo(const size_t offset) const;
const EdgeInfo* GraphTile::edgeinfo_with_cache(const size_t offset) const;
```

You can then return 2 different types, and a name in a method is more readable than an obscure `true` parameter in the calling site.

And with or without `unordered_map::reserve()`?

TeXitoi:
I've created `GraphTile::edgeinfo_with_cache()` and I reserve the cache at the first call to it.

kevinkreiser:
this seems like a good solution when all of the points are close together (high probability of cache hits). at this point all we need to do is have a reasonable idea of how often the slow down for cold cache will be a problem for the general use case.

kevinkreiser:
anyway im inclined to ship this but will run a quick benchmark of the scenario in which the points aren't all that close together. to see what performance does.

kevinkreiser:
ok so ive run some benchmarks of different points that i normally run to test loki.  first thing to note is that ive been running with the tiles as a memory mapped tar file. this has already given a decent improvement in both cached and uncached scenario. in fact they are almost the same at this point:

```
2016/10/03 12:11:37.626873 [INFO] Tile extract successfully loaded
2016/10/03 12:11:44.496228 [INFO] Succeeded Searches on Uncached Tiles
2016/10/03 12:11:44.496266 [INFO] --------------------------------
2016/10/03 12:11:44.496279 [INFO] Total: 3400
2016/10/03 12:11:44.496293 [INFO] Fastest: 40.344559,-79.998596 @ 0ms
2016/10/03 12:11:44.496303 [INFO] Slowest: 39.979069,-75.123100 @ 55ms
2016/10/03 12:11:44.496311 [INFO] Median: 2.375294ms
2016/10/03 12:11:44.496319 [INFO] Standard Deviation: 2.139621ms
2016/10/03 12:11:44.496326 [INFO] Faster Than 1 Standard Deviation: 5
2016/10/03 12:11:44.496333 [INFO] Slower Than 1 Standard Deviation: 157
2016/10/03 12:11:44.496340 [INFO] --------------------------------


2016/10/03 12:11:44.496544 [INFO] Failed Searches on Uncached Tiles
2016/10/03 12:11:44.496551 [INFO] --------------------------------
2016/10/03 12:11:44.496566 [INFO] No results
2016/10/03 12:11:44.496571 [INFO] --------------------------------


2016/10/03 12:11:44.496776 [INFO] Succeeded Searches on Cached Tiles
2016/10/03 12:11:44.496784 [INFO] --------------------------------
2016/10/03 12:11:44.496799 [INFO] Total: 3400
2016/10/03 12:11:44.496806 [INFO] Fastest: 40.344570,-79.998398 @ 0ms
2016/10/03 12:11:44.496813 [INFO] Slowest: 40.446320,-79.980400 @ 11ms
2016/10/03 12:11:44.496818 [INFO] Median: 2.213529ms
2016/10/03 12:11:44.496823 [INFO] Standard Deviation: 1.225102ms
2016/10/03 12:11:44.496828 [INFO] Faster Than 1 Standard Deviation: 8
2016/10/03 12:11:44.496834 [INFO] Slower Than 1 Standard Deviation: 480
2016/10/03 12:11:44.496839 [INFO] --------------------------------


2016/10/03 12:11:44.497017 [INFO] Failed Searches on Cached Tiles
2016/10/03 12:11:44.497023 [INFO] --------------------------------
2016/10/03 12:11:44.497038 [INFO] No results
2016/10/03 12:11:44.497043 [INFO] --------------------------------
```

i then applied this diff to baldr, and quickly patched my local version of loki to use it and i get:

```
2016/10/03 12:24:33.908755 [INFO] Tile extract successfully loaded
2016/10/03 12:24:41.873827 [INFO] Succeeded Searches on Uncached Tiles
2016/10/03 12:24:41.873863 [INFO] --------------------------------
2016/10/03 12:24:41.873875 [INFO] Total: 3400
2016/10/03 12:24:41.873888 [INFO] Fastest: 41.057171,-74.144798 @ 1ms
2016/10/03 12:24:41.873898 [INFO] Slowest: 39.979092,-75.123100 @ 54ms
2016/10/03 12:24:41.873906 [INFO] Median: 4.872647ms
2016/10/03 12:24:41.873913 [INFO] Standard Deviation: 2.713249ms
2016/10/03 12:24:41.873920 [INFO] Faster Than 1 Standard Deviation: 490
2016/10/03 12:24:41.873929 [INFO] Slower Than 1 Standard Deviation: 345
2016/10/03 12:24:41.873935 [INFO] --------------------------------


2016/10/03 12:24:41.874144 [INFO] Failed Searches on Uncached Tiles
2016/10/03 12:24:41.874151 [INFO] --------------------------------
2016/10/03 12:24:41.874166 [INFO] No results
2016/10/03 12:24:41.874171 [INFO] --------------------------------


2016/10/03 12:24:41.874371 [INFO] Succeeded Searches on Cached Tiles
2016/10/03 12:24:41.874378 [INFO] --------------------------------
2016/10/03 12:24:41.874394 [INFO] Total: 3400
2016/10/03 12:24:41.874401 [INFO] Fastest: 40.446659,-80.065598 @ 0ms
2016/10/03 12:24:41.874408 [INFO] Slowest: 40.664650,-74.217400 @ 11ms
2016/10/03 12:24:41.874414 [INFO] Median: 1.720000ms
2016/10/03 12:24:41.874419 [INFO] Standard Deviation: 1.112422ms
2016/10/03 12:24:41.874425 [INFO] Faster Than 1 Standard Deviation: 345
2016/10/03 12:24:41.874430 [INFO] Slower Than 1 Standard Deviation: 758
2016/10/03 12:24:41.874436 [INFO] --------------------------------


2016/10/03 12:24:41.874608 [INFO] Failed Searches on Cached Tiles
2016/10/03 12:24:41.874614 [INFO] --------------------------------
2016/10/03 12:24:41.874629 [INFO] No results
2016/10/03 12:24:41.874634 [INFO] --------------------------------
```

as noted earlier the uncached version is quite a bit slower, pretty much double.. i cant understand that.. the cached version is a bit faster but not double. its hard to say whether this is good or not... i'd hate to make the cold cache situation 100% slower to make the warm cache 30% faster without knowing what is more likely, warm or cold cache..

TeXitoi:
Can you explain "running with the tiles as a memory mapped tar file" ?

kevinkreiser:
yeah! its a new undocumented feature ;)

basically just take your tiles and do this:

`cd /my/tile/directory; find 0 1 2 3 | sort -n | tar cf planet.tar --no-recursion -T -`

then add this line to your valhalla.json config:

`"tile_extract": "/my/tile/directory/planet.tar",`

under the `mjolnir` section

TeXitoi:
Here, mmapped tiles doesn't change anything (is it working? it's not if I remove the `tile_dir` key). Note that my data are small: I only have the Paris region.

TeXitoi:
In the profiler (valgrind, which subestimate IO), most of the time is passed in decode7 and memory allocation for our test case.

kevinkreiser:
i cannot figure out where the extra time is.. i feel like we can make this work if we can just get that extra time on the cold cache down to no cache levels. its blowing my mind that its slower! i mean decode7 will happen more often in the old code vs your pr. so if its the dominant factor then you wouldnt expect cold cache to get slower since it does as many or less decode7 calls as the old non caching code. we must be missing something here..

kevinkreiser:
mmapped tars is enabled when you see some line like this: `2016/10/03 12:24:33.908755 [INFO] Tile extract successfully loaded`

TeXitoi:
Here, with mmaped tiles working, I have:
 - master mmaped: 3.94/3.81
 - this PR mmaped: 5.85/1.87
 - master no mmaped: 7.68/3.80
 - this PR no mmaped: 9.72/1.62

The overhead is in the unordered_map search: for each creation of the EdgeInfo object, you first search in the cache, which always miss the first time. But the cache is big, thus the search is not negligible.

TeXitoi:
I may have a better idea with no side effects (if it is effective, that's not sure). I'd like your view on the idea.

As a lot of time is passed in `shape()` in our case, and most of the time passed in shape is memory allocation, I propose 2 modifications:
 * `decode7` takes a `const boost::string_ref` instead of a `const std::string&`. That will remove the useless creation of a std::string in the call to `decode7()` in `edgeinfo()`.
 * add a `ForwardPointLLRange EdgeInfo::lazy_shape() const` that return a range of lazy points. Use it in the hot spot of `loki::Search`, i.e. in `project()`.

What do you think?

kevinkreiser:
@TeXitoi i really like this idea but i dont think we need to add boost as a dependency to midgard. if  you dont mind i can take a quick second to add a signature of the decode functions that takes a pointer and a count.

as for the lazy_shape, i am not sure i understand. do you mean do something like decode it as you go? meaning you get some subset of points each time you call `lazy_shape()`? sorry not quite understanding that bit!

TeXitoi:
I was thinking of a `boost::range` that compute the point when incrementing the iterator (so with the needed context). You would use it like that:
```c++
auto points = edgeinfo->lazy_shape();
if (points.empty() { return; }
PointLL prev = points.front();
points.drop_front();
for (const auto& cur: points) {
    // do the project method using prev and cur
    prev = cur;
}
```

Then you can lazily read the points without allocating, and decoding the byte stream inline in an abstract way. But if you don't want to depend on boost, we can do something similar by hand.

kevinkreiser:
wow, so i just did that change, and its even better:

```
2016/10/03 18:24:12.504314 [INFO] Tile extract successfully loaded
2016/10/03 18:24:19.161142 [INFO] Succeeded Searches on Uncached Tiles
2016/10/03 18:24:19.161176 [INFO] --------------------------------
2016/10/03 18:24:19.161192 [INFO] Total: 3400
2016/10/03 18:24:19.161207 [INFO] Fastest: 40.344589,-79.998901 @ 0ms
2016/10/03 18:24:19.161218 [INFO] Slowest: 39.979069,-75.123100 @ 61ms
2016/10/03 18:24:19.161227 [INFO] Median: 2.135882ms
2016/10/03 18:24:19.161237 [INFO] Standard Deviation: 2.108890ms
2016/10/03 18:24:19.161245 [INFO] Faster Than 1 Standard Deviation: 26
2016/10/03 18:24:19.161254 [INFO] Slower Than 1 Standard Deviation: 86
2016/10/03 18:24:19.161262 [INFO] --------------------------------


2016/10/03 18:24:19.161465 [INFO] Failed Searches on Uncached Tiles
2016/10/03 18:24:19.161475 [INFO] --------------------------------
2016/10/03 18:24:19.161483 [INFO] No results
2016/10/03 18:24:19.161490 [INFO] --------------------------------


2016/10/03 18:24:19.161709 [INFO] Succeeded Searches on Cached Tiles
2016/10/03 18:24:19.161720 [INFO] --------------------------------
2016/10/03 18:24:19.161729 [INFO] Total: 3400
2016/10/03 18:24:19.161740 [INFO] Fastest: 40.344589,-79.998901 @ 0ms
2016/10/03 18:24:19.161751 [INFO] Slowest: 40.727509,-74.167000 @ 24ms
2016/10/03 18:24:19.161760 [INFO] Median: 1.945882ms
2016/10/03 18:24:19.161769 [INFO] Standard Deviation: 1.268863ms
2016/10/03 18:24:19.161777 [INFO] Faster Than 1 Standard Deviation: 79
2016/10/03 18:24:19.161786 [INFO] Slower Than 1 Standard Deviation: 327
2016/10/03 18:24:19.161793 [INFO] --------------------------------


2016/10/03 18:24:19.161976 [INFO] Failed Searches on Cached Tiles
2016/10/03 18:24:19.161985 [INFO] --------------------------------
2016/10/03 18:24:19.161993 [INFO] No results
2016/10/03 18:24:19.162001 [INFO] --------------------------------
```

kevinkreiser:
@TeXitoi yeah a decode_iterator class would be awesome! i think we could add it to midgard and allow edgeinfo to return one. i would prefer we dont add boost but if its too complex without boost then maybe its worth it. please feel free to have a go at it!

TeXitoi:
I have a prototype working with an interface close to a priority queue, i.e. with `empty()`, `pop()` and `top()`. Promizing performances. Header only for inlining for the moment. Need to check a template API to provide the different Point types.

Then, edgeinfo returns a unique_ptr. Avoiding this allocation can be another interesting win.

kevinkreiser:
@TeXitoi Awesome sounds great!

TeXitoi:
With `lazy_shape()`, I'm at 2.2/2.2. PRs comming.

TeXitoi:
PRs: valhalla/midgard#104

I hope than, after this PR and removing allocation in `GraphTile::edgeinfo()`, we can close this PR as useless.

PR #299, opened 2016-10-03
dont copy the shape bytes into a string to decode them

kevinkreiser:
dont copy the encoded linestring when decoding it, this is an effort to provide an alternative to #298 as outlined by @TeXitoi


dnesbitt61:
:shipit:

PR #300, opened 2016-10-03
Updated error utility for internal errors

kdiluca:
- Added colons after Unknowns since these will most likely be followed by a exception.what() description

### DIFF
--- a/valhalla/baldr/errorcode_util.h
+++ b/valhalla/baldr/errorcode_util.h
@@ -131,7 +131,7 @@ namespace baldr {
     {181,"No suitable edges near location"},
     {182,"No data found for location"},
 
-    {199,"Unknown"},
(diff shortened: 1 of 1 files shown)


valhalla/baldr/errorcode_util.h
```diff
@@ -184,13 +184,13 @@ namespace baldr {
...
 
-    {499,"Unknown"},
+    {499,"Unknown:"},
 
     // tyr project 5xx
     {500,"Failed to parse intermediate request format"},
     {501,"Failed to parse TripDirections"},
 
-    {599,"Unknown"}
+    {599,"Unknown:"}
```
kevinkreiser: also is there really a need to okeep around the extra string? it just gets put into the message anyway

PR #305, opened 2016-10-07
Internal Error Code Updates

kdiluca:
- 	moved a few of the codes around, renumbered and removed colons from the unknowns
-	combined search errors into 1 error

### DIFF
--- a/valhalla/baldr/errorcode_util.h
+++ b/valhalla/baldr/errorcode_util.h
@@ -126,9 +126,7 @@ namespace baldr {
     {162,"Date and time is invalid.  Format is YYYY-MM-DDTHH:MM"},
     {163,"Invalid date_type"},
 
-    {170,"Locations are in unconnected regions. Go check/edit the map at osm.org"},
-    {171,"No suitable edges near location"},
-    {172,"No data found for location"},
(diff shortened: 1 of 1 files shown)


kdiluca:
@kevinkreiser , when you get a chance please look at this and then I'll submit all the PRs for the projects that will use it.

valhalla/baldr/errorcode_util.h
```diff
@@ -84,54 +84,51 @@ namespace baldr {
...
-    {164,"Path distance exceeds the max distance limit"},
-    {165,"Outside the valid walking distance at the beginning or end of a multimodal route"},
-    {166,"Outside the valid walking distance between stops of a multimodal route"},
-
-    {170,"Date and time required for origin for date_type of depart at"},
-    {171,"Date and time required for destination for date_type of arrive by"},
-    {172,"Date and time is invalid.  Format is YYYY-MM-DDTHH:MM"},
-    {173,"Invalid date_type"},
-
-    {180,"Locations are in unconnected regions. Go check/edit the map at osm.org"},
```
kevinkreiser: this shouldnt have been removed entirely. when the connectivity map check fails in loki we should be using this error not number 170 below

PR #307, opened 2016-10-18
be more relaxed in how you turn names into ids

kevinkreiser:
so if youre not careful loading up a tar extract of tiles wont work because of the paths. this makes the path to graphid much more relaxed in that it trims anything that isnt numbers or /'s. leaving only the level/id stuff that we should require. this could be more strict in that it could check the number of parts making up the id once we know the level but meh.


dnesbitt61:
:shipit:

PR #308, opened 2016-10-20
Branch 1.1.2

mtehver:
I am doing PRs for the changes I made to make Valhalla more mobile friendly. The changes are based on 1.1.2 release, so there could be some conflicts with the HEAD. Most changes are quite simple to help with discrepancies between C++ compiler/stdlib versions. I added support for compiling Valhalla with either boost::regex (the default) or std::regex (if USE_STD_REGEX preprocessor variable is set). I also abstracted graph storage handling into separate classes by refactoring filesystem based code from baldr GraphTile/GraphReader into abstract GraphTileStorage and GraphTileFsStorage. The storage handling code should perhaps include also write support (that would make mjolnir codebase cleaner), but that is not currently implemented.


kevinkreiser:
is there any way you can break this into smaller parts? the regex stuff is fine, and should be easily mergable. if you could make a PR for that against master that would be great.

the zip stuff doesnt seem to be included in the build itself so i dont think we're inclined to merge those until they build. they might need changes to the makefile and configure.ac. also they will be conflicting with some other changes we made to support memory mapped tar file loading. i wonder if you could possibly pull master and see what it would take to get these graphreader changes merged in with that.

mtehver:
I guess conflicts arise because I used 1.1.2 version as a base, not master. I had previously trouble compiling the projects when I used master branch, so decided to use official release instead.

I did not include the ZIP storage class in PR as this would require  external dependency with miniz library. The ZIP storage class I mentioned before was more like an experiment, in production we would probably use sqlite for storage (with similar structure to mbtiles, with gzipped tiles) or perhaps even custom storage class.

I will try to split the changes into multiple parts and use master as a base. Not sure when I am able to do this, though.

kevinkreiser:
sounds good, no worries! whenever you get around to it is fine with us

kevinkreiser:
we're going to close this one for now. hopefully you can pick it back up when you have a chance (many many things have changed since then!) let us know if you have any questions!

PR #309, opened 2016-10-26
make tiles more robust to future needs without breakeage

kevinkreiser:
so the trick here is that we put a bunch of unused spots for stuff in the header. the idea is that as we need to add more data into tiles we can use these slots for offsets into these new data structures. this wont break old versions as the slots will just be ignored. the only problem is that we cant use any of these slots for any information other than an offset,  we already have 48 spare bits in the header so there is a decent amount of room for growth there but i added another spare there just in case. 

all in all this will add 13mb to the planet tiles, which IMHO is nothing for the ability to safely add new chunks of data to the end of the tile.

thinking about it some more, you could actually use some of the empty slots for bitfields or unions but it couldnt be one that a previous version used as the end marker. so yeah even more flexible but have you be careful. we should develop version tests :smile:


dnesbitt61:
:shipit:

PR #313, opened 2016-11-08
Serialize the projected lat,lon as double.

dnesbitt61:
This avoids roundoff that was leading to shape issues.
fixes valhalla/loki#162


kevinkreiser:
this is crazy! great catch

dgearhart:
:ship:

PR #314, opened 2016-11-10
handle more complex representation from midgard gridded data

kevinkreiser:
this depends on valhalla/midgard#107 and essentially allows for polygons that have inners


dnesbitt61:
:shipit:

PR #316, opened 2016-11-15
Add edge merging

zerebubuth:
This adds a function `valhalla::baldr::merge` which takes a `GraphReader` and a functor, and iterates over all the edges in the graph to find those which are collapsible, calling the functor for each collapsed set of edges.

This is a building block for OpenLR-like descriptors, and intended to group edges into sequences where there are no true intersections, but there might be bridges or changes in other properties such as max speed.

@kevinkreiser could you take a look and let me know what you think, please?

### DIFF
--- a/src/baldr/merge.cc
+++ b/src/baldr/merge.cc
@@ -3,7 +3,6 @@
 #include "valhalla/midgard/logging.h"
 
 #include <boost/range/adaptor/map.hpp>
-#include <boost/optional.hpp>
 
 namespace bra = boost::adaptors;
 
@@ -215,8 +214,9 @@ struct edge_collapser {
   // returns the pair of nodes reachable from the given @node_id where they
   // are the only two nodes reachable by non-shortcut edges, and none of the
   // edges of @node_id cross into a different level.
-  boost::optional<std::pair<GraphId, GraphId> > nodes_reachable_from(GraphId node_id) {
-    boost::optional<GraphId> first, second;
+  std::pair<GraphId, GraphId> nodes_reachable_from(GraphId node_id) {
+    static const std::pair<GraphId, GraphId> none;
+    GraphId first, second;
 
     for (const auto &edge : iter::edges(m_reader, node_id)) {
       // nodes which connect to ferries, transit or to a different level
@@ -224,7 +224,7 @@ struct edge_collapser {
       if (edge.first->use() == Use::kFerry ||
           edge.first->use() == Use::kTransitConnection ||
           edge.first->trans_up() || edge.first->trans_down()) {
-        return boost::none;
+        return none;
       }
 
       // shortcut edges should be ignored
@@ -237,13 +237,13 @@ struct edge_collapser {
       // needed at this node to make it collapsible, or the node has more than
       // two edges.
       if (m_tracker.get(edge.second)) {
-        return boost::none;
+        return none;
       }
 
       if (first) {
         if (second) {
           // can't add a third, that means this node is a true junction.
-          return boost::none;
+          return none;
 
         } else {
           second = edge.first->endnode();
@@ -254,13 +254,13 @@ struct edge_collapser {
     }
 
     if (first && second) {
-      return std::make_pair(*first, *second);
+      return std::make_pair(first, second);
     } else {
-      return boost::none;
+      return none;
     }
   }
 
-  boost::optional<GraphId> next_node_id(GraphId last_node_id, GraphId node_id) {
+  GraphId next_node_id(GraphId last_node_id, GraphId node_id) {
     //
     //        -->--     -->--
     //   \   /  e4 \   /  e1 \   /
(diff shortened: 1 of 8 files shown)


src/baldr/merge.cc
```diff
@@ -0,0 +1,414 @@
...
+namespace merge {
+
+namespace {
+
+// a place we can mark what edges we've seen, even for the planet we should
+// need ~ 100mb, as it allocates one bit per ID. there are currently around
+// 453 million ways in OSM, and we might allocate two edge IDs per way,
+// giving something like 108mb of bits needed.
+struct bitset_t {
+  bitset_t(size_t size) {bits.resize(size);}
```
kevinkreiser: there were some bugs with that bitset thing where it took up more memory than it needed: https://github.com/valhalla/tools/commit/d85fd7ab21087e03ed9470e0148bcd47d514b778#diff-5d0751e2f6630073e4f54144251387f0

src/baldr/merge.cc
```diff
@@ -0,0 +1,414 @@
...
+  }
+  return tile_count;
+}
+
+edge_tracker::edge_index_t edge_tracker::edges_in_tiles(GraphReader &reader) {
+  uint64_t tiles_in_levels = count_tiles_in_levels(reader);
+
+  //keep the global number of edges encountered at the point we encounter each tile
+  //this allows an edge to have a sequential global id and makes storing it very small
+  LOG_INFO("Enumerating edges...");
```
kevinkreiser: we shouldnt log in the lower level stuff we can just move logging to the calling application. same for log below

src/baldr/merge.cc
```diff
@@ -0,0 +1,414 @@
...
+        //
+        // TODO: presumably this is bad when the cache doesn't behave like a
+        // cache and evict older tiles. but this shouldn't be a problem when
+        // mmapping the whole extract.
+        //reader.Clear();
+      }
+    }
+  }
+
+  return std::move(edges_in_tiles);
```
kevinkreiser: is the explicit `move` necessary? i think the compiler is required to do the move for you in this case

src/baldr/merge.cc
zerebubuth: Ooops, yeah. Not enough to have noticed on the small cases I've been testing on. Thanks for pointing it out, fixed in 0e1da23.

src/baldr/merge.cc
zerebubuth: Yup, good call. Fixed in 6cda672.

src/baldr/merge.cc
```diff
@@ -0,0 +1,414 @@
...
+      GraphId tile_id{i, level.level, 0};
+      if (reader.DoesTileExist(tile_id)) {
+        //TODO: just read the header, parsing the whole thing isnt worth it at this point
+        edges_in_tiles.emplace(tile_id, edge_count);
+        const auto* tile = reader.GetGraphTile(tile_id);
+        edge_count += tile->header()->directededgecount();
+        // this clears the cache, and the test relies on being able to inject
+        // stuff into the cache :-(
+        //
+        // TODO: presumably this is bad when the cache doesn't behave like a
```
zerebubuth: Ah, I had missed the `GraphReader::OverCommitted()` function. Thanks for pointing it out. Fixed in 709ec6e.

src/baldr/merge.cc
```diff
@@ -0,0 +1,432 @@
...
+  }
+
+  const auto* tile = reader.GetGraphTile(max_tile_id);
+  edge_count += tile->header()->directededgecount();
+  return bitset_t(edge_count);
+}
+
+namespace iter {
+
+// edges provides a container wrapper for the edges leaving a node which is
```
kevinkreiser: word `edges` at the beginning might have meant to be removed?

src/baldr/merge.cc
zerebubuth: Yup, that seems to be [the right way to do it](http://stackoverflow.com/questions/17473753/c11-return-value-optimization-or-move/17473874#17473874). I find the interaction between return value optimisation and move semantics to be quite confusing. Fixed in ffea013.

src/baldr/merge.cc
```diff
@@ -0,0 +1,432 @@
...
+    : m_reader(reader)
+    , m_tracker(tracker)
+    , m_func(func)
+    {}
+
+  // returns the pair of nodes reachable from the given @node_id where they
+  // are the only two nodes reachable by non-shortcut edges, and none of the
+  // edges of @node_id cross into a different level.
+  boost::optional<std::pair<GraphId, GraphId> > nodes_reachable_from(GraphId node_id) {
+    boost::optional<GraphId> first, second;
```
kevinkreiser: instead of `optional<GraphId>` could we implement `GraphId::operator bool() const` and have it return `GraphId::Is_Valid()` or something?

kevinkreiser:
after boost optional comment above :rocket:

zerebubuth:
I removed optional in commit a8ac5ad, and added `GraphId`'s conversion to bool in c35edf7. That seems to build okay across the whole of Valhalla, so hopefully no unexpected changes in behaviour. It uses the new "implicit explicit" conversions, so should only happen implicitly for use in `if` statements and similar.

Added more flexibility in f923e32 and b360b75, which allow the customisation of the `GraphTile`s to iterate over and the edges to consider for collapsing, respectively. These can now be passed in as a range object and `std::function<>`, allowing the client code to decide the policy.

@kevinkreiser let me know what you think?

PR #320, opened 2016-11-22
Added Location::ToPtree

dgearhart:


kevinkreiser:
im looking at makign this not needed

kevinkreiser:
:rocket

PR #322, opened 2016-11-28
fix includes but also start towards conditional usage of boost regex

kevinkreiser:
previously merged pr slightly broke things in that it didnt export all of the headers. ive also started towards making using regex from std lib easy via configure but it doesnt work right just yet. somehow it still tries to link boost regex (at least on linux)


dgearhart:
:shipit:

dgearhart:
:ship: when it works

PR #323, opened 2016-11-30
Explicitly set end_offset_ as its own data member 

dnesbitt61:
rather than an entry in the empty_slots array. Turns out that if you add new entries the slot in empty_tiles
used to set the end offset changes position in the header, so an old tile end offset is going to be 0. This makes sure the end offset is always at the end of the graph tile header.


gknisely:
:ship:

PR #324, opened 2016-12-01
Add elevation data to Valhalla SW-87776

noblige:
Adding elevation data per edge cc @dnesbitt61 @kevinkreiser 
https://github.com/valhalla/skadi/pull/65
https://github.com/valhalla/baldr/pull/324
https://github.com/valhalla/mjolnir/pull/573


noblige:
@kevinkreiser I'm also considering about adding another 64bit spare and storing elevation at full resolution (14 bits). Thoughts?
cc @yzadik

dnesbitt61:
@noblige  We just went through a tile redesign effort with Valhalla 2.0 to make it so we can maintain compatibility yet allow for some expansion. To maintain compatibility, none of the existing structure sizes can change (nor can positions of fields within structures), so adding 64 spare bits to DirectedEdge is undesirable. There are several options that we prefer for adding mean_elevation (prefer that name if it applies to an entire edge):

1) If NOT anticipated to be used in costing but rather just returned as an attribute, the best choice may be to add it to EdgeInfo (which has 12 spare bits). This has the advantage of not being stored for each edge direction, leading to smaller growth in tile sizes.

2) An "empty" slot can be used to store an array of short int with 1:1 correlation to the directed edge index. We can help with adding new structures to a tile. These values can then be indexed using the DirectedEdge graph Id to get mean elevation for attribution. This is also best used if the mean elevation is primarily used as an attribute returned via trace_attributes. However, it could be used in costing where needed.

3) If you envision other elevation based fields that may need to be used in costing it may make sense to create a new Elevation structure and store this as an array in one of the empty slots. We could possibly move the weighted_grade, max_slopes, mean_elevation, and perhaps new fields into a single (64 bit?) structure. This could then be access by costing methods that need it, and not read by costing methods that do not.

I prefer keeping the current DirectedEdge spare bits available for attributes that might be needed by a majority of costing methods.

noblige:
@dnesbitt61 thanks for the feedback, it makes sense. Option 3 would probably be most flexible, as this will allow to store not just mean elevation, but multiple elevation/chainage pairs (for longer edges). This will possibly make other (slope, grade) fields redundant, as they can be computed in real-time.
I will close this PR for now.

kevinkreiser:
i also quite like option 3 as its very flexible and also allows a nicer logical grouping.

PR #330, opened 2016-12-15
Add named flag. 

dnesbitt61:
Separate spare from the endnode so we can actually use the spare bits in GraphId.


dgearhart:
:shipit:

PR #331, opened 2016-12-15
Merged master and fixed a bug

zerebubuth:
NOTE: this is a merge into the segments branch, not master.

The loop which was exploring edges was testing the mark on "seen" edges, which meant that it would see the mark it just made and assume that the edge could not be explored. Checking the mark is necessary before starting to explore a node, to ensure that edges are not explored twice. However, in the exploration loop it is not required, and prevents segments of longer than length 2 being created.

I've added a simple test case for this as well.

@dnesbitt61 could you review, please?


dnesbitt61:
:shipit:

PR #333, opened 2017-01-04
Add traffic segment associations

dnesbitt61:
Adds support for associating directed edges to traffic segments. Each edge can associate to a single traffic segment, with the segment Id, percent along the segment at the start and at the end of the edge, and flags indicating whether the edge starts a traffic segment and/or ends a traffic segment. For now support for cases where a directed edge associates to multiple traffic segments is not complete. This data will be stored in "chunks" which indicate the traffic segment Ids and weighting to apply to each.


kevinkreiser:
:+1:

PR #335, opened 2017-01-09
Return nullptr from GetGraphTile if lat,lon is out of range

### LINKED ISSUE #334: Out-of-bound coordinates results in invalid tile id
When input coordinates are out of longtitude,latitude bounds, meili says it gets invalid tile ids.
```
6 2017/01/08 11:23:14.297012 POST /mm/?mode=multimodal&search_radius= HTTP/1.0
2017/01/08 11:23:14.297466 [INFO] Got Map Matching Request 6
2017/01/08 11:23:14.297685 [ERROR] src/prime_server.cpp:430 worker_t: Tile id out of valid range
```

Diving into the code, I found that errors were reported here https://github.com/valhalla/meili/blob/master/src/meili/candidate_search.cc#L199-L200

Maybe baldr should do some work to normalize out-of-bound coordinates to [-180,180] and [-90,90], or at least let `GetGraphTile` return nullptr when it sees out-of-bound coordinates?

dnesbitt61:
Since tile extent can be configurable it doesn't make sense to normalize lat,lon if out or normal ranges.

Fixes #334

### DIFF
--- a/src/baldr/graphreader.cc
+++ b/src/baldr/graphreader.cc
@@ -135,7 +135,8 @@ const GraphTile* GraphReader::GetGraphTile(const GraphId& graphid) {
 }
 
 const GraphTile* GraphReader::GetGraphTile(const PointLL& pointll, const uint8_t level){
-  return GetGraphTile(tile_hierarchy_.GetGraphId(pointll, level));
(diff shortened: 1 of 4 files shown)


kevinkreiser:
:rocket:

PR #336, opened 2017-01-13
Add bounding-box based "GetTiles" to TileHierarchy

zerebubuth:
Add a couple of convenient functions to `TileHierarchy` to return the IDs of tiles which intersect given bounding boxes at some or all levels.

@kevinkreiser could you review, please?

### DIFF
--- a/src/baldr/tilehierarchy.cc
+++ b/src/baldr/tilehierarchy.cc
@@ -47,5 +47,24 @@ uint8_t TileHierarchy::get_level(const RoadClass roadclass) const {
   }
 }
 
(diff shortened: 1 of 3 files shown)


kevinkreiser:
i might have used lists instead of vectors (since the caller probably doesnt need a random access container) so the second function could splice the results of the various levels together but it doesnt much matter i dont think!

:rocket:

zerebubuth:
I thought about `std::list`, but figured that a smaller number of large allocations for the `std::vector`s was probably better than the large number of small allocations in the list. I might be wrong! If I am, then I'm sure we'll find it when profiling later on :wink:

PR #337, opened 2017-01-21
add lane connectivity information

noblige:
publishing for early feedback cc @kevinkreiser @dnesbitt61


dnesbitt61:
This looks good to me!

PR #340, opened 2017-01-23
Return nullptr from GetGraphTile if the tile Id is invalid

### LINKED ISSUE #338: GetGraphTile should return nullptr if graphid is invalid
```c++
const GraphTile* GraphReader::GetGraphTile(const GraphId& graphid);
```

invalid graphid will trigger:

```
2017/01/22 17:36:58.100516 [ERROR] src/prime_server.cpp:430 worker_t: Could not compute FileSuffix for non-existent level
```

dnesbitt61:
Fixes #338


kevinkreiser:
:ship:

PR #342, opened 2017-01-29
Updates to TrafficAssociation structures

dnesbitt61:
Use less memory for the most common case: a 1:1 edge to segment association where the segment is in the same tile as the edge. Also update the TrafficChunk structure. Add a new structure, TrafficSegment, to return from the graph tile.

### DIFF
--- a/valhalla/baldr/trafficassociation.h
+++ b/valhalla/baldr/trafficassociation.h
@@ -6,10 +6,10 @@
 namespace valhalla {
 namespace baldr {
 
-// 6 bits are used for percentages and weights in traffic associations
-constexpr uint64_t kEndsSegment   = 63;
-constexpr float kPercentFactor    = 63.0f;
-constexpr float kInvPercentFactor = 1.0f / 63.0f;
+// 8 bits are used for percentages
+//constexpr uint64_t kEndsSegment   = 255;
+constexpr float kPercentFactor    = 255.0f;
+constexpr float kInvPercentFactor = 1.0f / 255.0f;
 
 // These are used to indicate a chunk is stored for this edge's
 // association to traffic segments. A chunk count and chunk index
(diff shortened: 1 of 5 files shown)


valhalla/baldr/trafficassociation.h
```diff
@@ -6,123 +6,219 @@
...
 namespace baldr {
 
-constexpr float kPercentFactor    = 127.0f;
-constexpr float kInvPercentFactor = 1.0f / 127.0f;
-
-// TODO - need to redesign the traffic chunks...
-
-// These are used to indicate
+// 8 bits are used for percentages
+//constexpr uint64_t kEndsSegment   = 255;
```
kevinkreiser: the code below uses this but its commented out, is it somwhere else?

valhalla/baldr/trafficassociation.h
```diff
@@ -6,123 +6,211 @@
...
+
+// Structure used to return information
+struct TrafficSegment {
+  GraphId segment_id_;
+  float begin_percent_;
+  float end_percent_;
+  bool starts_segment_;
+  bool ends_segment_;
+
+  TrafficSegment(const GraphId& id, const float begin_pct,
```
kevinkreiser: why does this need a constructor?

valhalla/baldr/trafficassociation.h
```diff
@@ -6,123 +6,211 @@
...
 
-// TODO - need to redesign the traffic chunks...
+/**
+ * TrafficChunk describes many to one and many to many associations of
+ * traffic segments to edges. They are also used for cases where a segment
+ * is not within the same tile as the edge (at tile boundaries or if segment
+ * is associated to an edge on a different hierarchy level). There can be
+ * multiple chunks per edge.
+ */
+class TrafficChunk {
```
kevinkreiser: can we make this a struct and take away all the superfluous functions?

kevinkreiser: oh i guess we need the constructor because of the `kPercentFactor`?

kevinkreiser:
im not a huge fan of having structs act like classes for no benefit but theres nothing really wrong with it if we want to keep that style in our code (since these are likely to be inlined anyway?)

kevinkreiser:
i guess im saying ship it if you dont feel like changing it haha

valhalla/baldr/trafficassociation.h
dnesbitt61: I made this a class since some data should not be accessed directly (conversion to float percent from an internal int value for example)

